PR #16, opened 2015-04-25
Suggestion : Change icon to paste

RaphaelRochet:
This is really really subjective matter but ... how about using the "paste" icon instead of "cut" on the clipboard indicator ? It's a little closer to the extension job, which is all about what we want to paste, and the paste symbolic icon has a nice clipboard look too. Or maybe the "cut" has some special meaning for you ?

We may also go further and add a preference where one can choose between cut/copy/paste icons. 

Edit :
In case you wonder what is the 'paste' icon, here it is :
screenshot from 2015-04-25 22-42-31


danielquinn:
+1

Tudmotu:
OK.

PR #18, opened 2015-04-26
menu: put latest element on top

erick2red:
If you keep a number of commit too large, the menu grows too big and you can see with a click which is the last you have in there. By reverting the order and placing latest elements on top, you fix this situation.


Tudmotu:
Yes! Thank you :)

PR #49, opened 2016-09-07
Toggle notification on copy

optama:
Option in gsettings schema and settings panel to toggle showing of notification on copy


Tudmotu:
Thank you @optama ! :smile:

PR #52, opened 2016-10-12
Spanish translation

junquera:
There is the spanish translation. I'm not sure if the .mo file is well done, I've never worked with locales and I don't know well the files format, but .po file is ok.


Tudmotu:
Thanks! 😃

PR #69, opened 2017-07-09
Fix 'Toggle the menu' function

carmanaught:
Update the _toggleMenu function to use .menu.toggle() as the existing menu visibility check does not work.


Tudmotu:
Thanks! :smiley: 
I will upload it to e.g.o soon

PR #70, opened 2017-07-31
make max clipboard length 1000

DwordPtr:


Tudmotu:
Thanks. I guess 50 _is_ a bit low, and if someone wants more they can now do that.

PR #74, opened 2017-08-12
Fix signal handler disconnect

carmanaught:
This fixes the disconnection of the the signal handler when removing old entries which was throwing gsignal instance has no handler with id errors.


Tudmotu:
Thanks! :)

PR #78, opened 2017-09-16
Option to disable cache (registry.txt) file

JPvRiel:
Tested:

- When disabled is set, no new entries are created in registry.txt
- When disabled is not set, entries are created in registry.txt

There are some subtle quirks on edge cases, but otherwise functional:
- Old stale entries in registry.txt will remain (disabling does not delete registry file)
- When deleting an old/stale item stored before disabling, if the item was previously in registry.txt, the delete will only remove it from memory, but not from registry.txt
- If a registry file previously existed, and was disabled, and then is re-enabled, the previous registry file is overwritten and replaced with the current clipboard content.
- If gnome-shell is restarted, the old stale registry.txt file gets loaded (this behaviour is probably undesirable)

A future tweak might be to explicitly trigger deleting the registry file when toggled to disable in order to clear up the quirks. For now, if users really want this feature, suggest deleting the registry.txt file manually.


JPvRiel:
Just a nude here to see if this looks any good for a merge?

PR #88, opened 2017-10-27
Feature: display current clipboard content in topbar 

sVerentsov:
Adds a feature of displaying current clipboard content in topbar. 
The layout of topbar (icon, clipboard or both) and number of displayed characters can be adjusted in settings.

Display clipboard content in top bar:
image
Display content+icon:
image

In private mode shows ellipsis instead of clipboard content (if enabled):
image


Tudmotu:
Hi @sVerentsov,
Thanks for putting an effort into this!
Nice feature :smiley: 
I don't mind adding this functionality, but you should make the default be just the icon, as it is now. Also, please extract the ``if (DISPLAY_MODE == x) ...`` to a method and use it in both places.
I will test it after those changes are made. Thanks! :)

sVerentsov:
@Tudmotu Fixed.

Tudmotu:
Nice, thanks! :)
I will check it out soon

PR #90, opened 2017-10-28
Implementation of a basic search feature

Maestroschan:
It's not perfect but it's simple and it works :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,12 +10,12 @@ https://extensions.gnome.org/extension/779/clipboard-indicator/
 Installation
 ----------------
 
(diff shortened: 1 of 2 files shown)


Tudmotu:
Nice!! :smiley: 
Gonna test it and hopefully merge soon. Thanks!! :)

Maestroschan:
As i explained in the commit for extension.js, there are some problems with this implementation, which can be solved easily but it would mean more complexity, i didn't know if it is a problem for you so i let it like this.

This pull request also add a string to translate, which i totally forgot to add to the .pot file

PR #96, opened 2017-11-03
Updated Czech translation

AsciiWolf:


Tudmotu:
Nice, thanks!! :)

AsciiWolf:
@Tudmotu I see you did not merge it yet. Is there any problem with my PR?

Tudmotu:
Nope, just forgot to click it :wink:

AsciiWolf:
@Tudmotu It looks like that the "Type here to search...", "Icon", "Clipboard Content" and "Both" translations are not applied for some reason. Do you have any idea what could be the issue?

AsciiWolf:
That's weird, "Type here to search..." is now correctly translated after a reboot, but the other strings are still untranslated.

Tudmotu:
Thanks, @AsciiWolf - seems like these are not actually using the gettext function:
https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/blob/master/prefs.js#L205

PR #104, opened 2017-11-29
code quality changes, simple build script

dmig:
I'll start my series of pull requests with this one. It doesn't change anything, just addresses overall code quality:
- fix 1 console warning about undefined variable
- fix bunch of linter warnings
- add ESLint config for code checking

Also I added simple script to update translation template, translation files from new template and compile setting schema.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,4 @@
+module.exports = {
+    plugins: ['gjs'],
+    extends: ['plugin:gjs/extension']
+};
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+xgettext -o clipboard-indicator.pot extension.js prefs.js utils.js
+find locale/ -name \*.po -exec msgmerge -U {} clipboard-indicator.pot \;
+glib-compile-schemas schemas/
--- a/extension.js
+++ b/extension.js
@@ -20,10 +20,10 @@ const _ = Gettext.domain('clipboard-indicator').gettext;
 const Clipboard = St.Clipboard.get_default();
 const CLIPBOARD_TYPE = St.ClipboardType.CLIPBOARD;
 
-const SETTING_KEY_CLEAR_HISTORY = "clear-history";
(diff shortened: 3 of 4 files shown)


Tudmotu:
Hi @dmig, thank you very much for the effort and this PR! :smiley: 
Unfortunately, I prefer not incorporating of these changes. ESList is not something I want to use, and the stylistic changes to the code are partly nice and partly not my taste.

I really like the ``build.sh`` script you added, though I would rename it something more appropriate (like ``build-translations.sh``).

If you could open a PR for the build script on its own (and rename it), that would be awesome :grin:
I can also just cherry-pick it, if you don't feel like making another PR.

And again, thanks for the effort!

dmig:
These "stylistic" changes are for code to comply to official style guide: https://wiki.gnome.org/Projects/Gjs/StyleGuide
You may use ESLint or may not, but it's config will allow anyone to keep same coding style.

dmig:
Current `build.sh` is a bit dirty solution, it will update `.pot` and `.mo` files on every run. So I consider it's a temporary tool. It'd be better to check if `.pot` is changed in any line other than `POT-Creation-Date`. 
Also `.po` and `gschemas.compiled` may be excluded from repository, as they are generated.

Tudmotu:
Yeah, the main issue with "style guides" and tools like ESlint is that they inevitably cause discussions on what is the "right" way to write code. I believe every way is "right", and I don't want to limit people to a specific style. If you would like discussing this further and hear more on my views, we can do this via email :)

Regarding the "compile .pot file every build" - that's not necessarily bad. Nowadays every change causes the translations to move from one line to another, so it's even appropriate to update the translation before every release. We can also add a ``build.sh`` script that runs the ``build-translations.sh`` and then ``bundle.sh`` to prepare for release. And in the utopian future where we have gnome-shell tests, it will also run ``run-tests.sh`` :smile:

dmig:
I think as far as we develop for Gnome Shell, we should follow it's developers guides. That's all.
Especially when they provide reasonable guides.

Tudmotu:
I am closing this PR. I might cherry pick that script some when tho :)

PR #106, opened 2017-12-02
Update polish translation

pkomur:


locale/pl/LC_MESSAGES/clipboard-indicator.po
```diff
@@ -7,70 +7,107 @@ msgid ""
...
 "Content-Transfer-Encoding: 8bit\n"
-"X-Generator: Poedit 1.8.9\n"
+"X-Generator: Poedit 2.0.4\n"
 "Plural-Forms: nplurals=3; plural=(n==1 ? 0 : n%10>=2 && n%10<=4 && (n%100<10 "
 "|| n%100>=20) ? 1 : 2);\n"
 
-#: extension.js:113
+#: extension.js:78
+msgid "Text will be here"
+msgstr ""
```
Tudmotu: I think you might forgot to translate this string?

pkomur: I have not forgotten about it. I just don't understand, what does it exactly mean.

Tudmotu: Ah ok, you are right, this text is weird.
From the code, I can see that this is the initial label that the indicator button has in the topbar, if the "show clipboard text" option is on. But it is immediately replaced with ``...`` so I don't think it's an issue.

Thank you for this PR! :smiley:

PR #115, opened 2018-03-10
Update Spanish translation; add Catalan translation

fitojb:
I moved the existing Spanish translation from `es_ES` to just `es`, because it doesn’t have any Spain-specific lingo; then I refreshed it from the latest POT. The new Catalan translation is in a separate commit.


Tudmotu:
Thanks!

PR #132, opened 2018-08-10
Tweaked Czech translation a bit

p-bo:


locale/cs/LC_MESSAGES/clipboard-indicator.po
```diff
@@ -13,21 +13,20 @@ msgstr ""
...
 
 #: extension.js:119
 msgid "Type here to search..."
-msgstr "Vyhledávejte psaním sem..."
+msgstr "Vyhledávejte psaním sem…"
 
 #: extension.js:166
 msgid "Private mode"
-msgstr "Soukromý mód"
+msgstr "Stop (proti úniku citlivých údajů)"
```
AsciiWolf: I think the current (literal) translation is much better.

locale/cs/LC_MESSAGES/clipboard-indicator.po
```diff
@@ -59,27 +58,27 @@ msgstr "Následující položka"
...
 
 #: prefs.js:134
 msgid "Max cache file size (kb)"
-msgstr "Maximální velikost mezipaměti (kb)"
+msgstr "Omezit velikost mezipaměti na nejvýše (KB)"
 
 #: prefs.js:139
 msgid "Disable cache file"
-msgstr "Zakázat soubor mezipaměti"
+msgstr "Nepoužívat mezipaměť (soubor)"
```
AsciiWolf: "Nepoužívat soubor mezipaměti" sounds better.

locale/cs/LC_MESSAGES/clipboard-indicator.po
```diff
@@ -59,27 +58,27 @@ msgstr "Následující položka"
...
+msgstr "Interval obnovování (ms)"
 
 #: prefs.js:129
 msgid "Preview Size (characters)"
 msgstr "Velikost náhledu (znaky)"
 
 #: prefs.js:134
 msgid "Max cache file size (kb)"
-msgstr "Maximální velikost mezipaměti (kb)"
+msgstr "Omezit velikost mezipaměti na nejvýše (KB)"
```
AsciiWolf: What's wrong with the lowercase kb?

AsciiWolf:
Also, if possible, please add the "Cache only favorites" string that is currently missing in the po file. Otherwise LGTM.

Tudmotu:
Thank you for reviewing this @AsciiWolf!
If you have time to fix these issues and resubmit this PR I would greatly appreciate it :slightly_smiling_face: 
You can create a new branch on your fork, merge [their branch](https://github.com/p-bo/gnome-shell-extension-clipboard-indicator/tree/patch-1) and resubmit a new PR.

AsciiWolf:
@Tudmotu I think I'll wait for @p-bo to fix this instead of forking his work. :-)

locale/cs/LC_MESSAGES/clipboard-indicator.po
p-bo: I do see from point of view of someone, who already know, what is this function doing and is bored to read that long string. But what about users, who wasn't use this function before? For example, in Firefox there is also similar situation with private mode, but when someone use it, (s)he gets explanation in opened window/tab, so user is not lost and know exactly, what is going on. I tested furthermore with users and result is, that it could be maybe even better to replace word "Stop" with "Pozastavit" (space wise, it fits into menu without problems). Even more explanatory would be: "Pozastavit (proti nechtěnému vložení citlivých údajů)". See, Gnome as whole is intended for wide user base (and mainly at not so experienced users), so be as explanatory as is possible is good thing (see ease of comprehensives of strings for example in Android and other consume platforms).

p-bo:
@Tudmotu @AsciiWolf regarding "Cache only favorites" – reported in bugreport #134 too…

locale/cs/LC_MESSAGES/clipboard-indicator.po
p-bo: Lowercase b is not correct for this case – kilobit vs kilobyte:
https://en.wikipedia.org/wiki/Kilobit
https://en.wikipedia.org/wiki/Kilobyte
correct form would be kB, but it is practically not used anywhere, most common is using prefix as upper case letter. So KB, MB, GB (or KiB, MiB, GiB for binary).

locale/cs/LC_MESSAGES/clipboard-indicator.po
p-bo: It is don't use what? Cache. Detail – implemented as what? File.
@Tudmotu what is the point of communication with user there?  Is it about store data only in RAM (so available only during desktop session) instead in file (where it persist across sessions)?
In that case it could be there "Don't save to disk (no persistence)" maybe?
That "Disable cache file" may be too much technical for average user (as kind mainly targeted by Gnome project)…

Issue #143, opened 2018-10-17
[Feature Request] - Configuring the movement of items to the top of the list when selected.


damjad:
Hi!

I hope you are fine.

I've been a Mac user for the last couple of years and made a shift towards Linux based systems. (Yeah! I'm a noobie :) ). I've been using [FlyCut](https://github.com/TermiT/Flycut) on OSX. According to them, it's a clipboard made for developers. It's a very nice and simple clipboard manager like this one.  

The only difference is that this one does not retain the position of a selected item in the history and moves the selected item to the top. While copying and pasting code snippets from history, it's hard to keep track of the position of the items, especially when copying a list of items in a certain order and pasting them in the same order.

I've tried to fix the issue myself. Here is the [link](https://github.com/damjad/gnome-shell-extension-clipboard-indicator).

I'll be really glad to hear from you about any suggestions or help.

PR #144, opened 2018-10-17
Configured the movement of items to the top of the list when selected

damjad:
Tried to solve #143.


Tudmotu:
Hi @damjad ! Thanks! :slightly_smiling_face: 
I also dislike the current behavior. It also messes up the shortcuts (cause the selected item keeps jumping).

I will check this out hopefully somewhat soon and will probably merge it. Thanks!

schemas/org.gnome.shell.extensions.clipboard-indicator.gschema.xml
```diff
@@ -77,6 +77,14 @@
         </description>
     </key>
 
+    <key name="move-item-first" type="b">
+      <default>true</default>
```
Tudmotu: Can you please change the default value to ``false``?
I would like this to be the default behavior.

damjad: Thank you for reviewing. I've changed it to `false`.

PR #146, opened 2018-11-04
add strip text to clipboard indicator

aliva:


Tudmotu:
This has been merged but not marked as such due to github bug.

Issue #154, opened 2019-04-18
Expand history size to 9999 (Or at least more than 500)


luisalvarado:
Currently the hard history limit of 50 is too small. As a developer, getting to 50 can take just less than 10 minutes and I might need something I copied an hour ago, or yesterday which is typically the case. Raising the limit to 9999 or something higher than 500 would help at many levels.

SUPERCILEX:
Duplicate of #222

PR #155, opened 2019-04-30
Allow setting history size up to 200 in settings.

### LINKED ISSUE #102: History size constraints
Hi!
Can you just increase the constraint for the history size, from 50 to, let's say, 100?
Sometimes 50 is just not enough for me, and I had to change this constraint locally in your scripts.

Thanks. :)

BlueBoxWare:
Fixes #102
Follow up to ac96d1a591bd423ed986e3719efc39e58de9e3ed

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -61,7 +61,7 @@ const App = new Lang.Class({
         this.field_size = new Gtk.SpinButton({
             adjustment: new Gtk.Adjustment({
                 lower: 1,
-                upper: 50,
+                upper: 200,
                 step_increment: 1
             })
         });


nlpsuge:
#154 

Hi, 200 maybe is enough for the most people, but too small for busy developers.

nlpsuge:
@BlueBoxWare 

Are you true you can set a value greater than 50? I tried to set value to 51(or greater), but it will change to 50 after reopening the Setting dialog.

Any advice?

BlueBoxWare:
> @BlueBoxWare
> 
> Are you true you can set a value greater than 50? I tried to set value to 51(or greater), but it will change to 50 after reopening the Setting dialog.
> 
> Any advice?

What is the output of

`gsettings range org.gnome.shell.extensions.clipboard-indicator history-size`?

nlpsuge:
@BlueBoxWare 
The output is:
No such schema “org.gnome.shell.extensions.clipboard-indicator”

BlueBoxWare:
And the file `prefs.js` (probably in `~/.local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com/`) in your local copy of clipboard indicator contains this:

```javascript
        this.field_size = new Gtk.SpinButton({
            adjustment: new Gtk.Adjustment({
                lower: 1,
                upper: 200, // this should be 200 or 9999 or whatever you prefer
                step_increment: 1
            })
        });
```

nlpsuge:
I'v tried your method. But not works for me. I don't know why...so strange...

BlueBoxWare:
Hmm, yeah that's weird. I have no idea what the issue could be.

nlpsuge:
I can set History Size up to 200 now. Thanks.

nlpsuge:
@BlueBoxWare 
Hi, 

After modifying ```history size``` stored in ```prefs.js``` and ```schemas/org.gnome.shell.extensions.clipboard-indicator.gschema.xml```, I run ```glib-compile-schemas schemas/``` to generate a brand-new ```gschemas.compiled```. After rebooting ```gnome-shell```, now I can change the history size in the UI.

So can I say ```gschemas.compiled``` is used by gnome-shell extension at the run-time other than ```schemas/org.gnome.shell.extensions.clipboard-indicator.gschema.xml```?

BlueBoxWare:
@nlpsuge
I think so, but I'm not 100% sure how those Gnome schema's actually work.

nlpsuge:
@BlueBoxWare 
I learn something from this link:
https://askubuntu.com/questions/1178580/where-are-gnome-extensions-preferences-stored

```The default extension settings are located in an *.xml file```, in this case, should be ```schemas/org.gnome.shell.extensions.clipboard-indicator.gschema.xml```.

```If you edit these files on your installed system, you will need to recompile the "gschema" by running one of the following commands```, that is, ```glib-compile-schemas schemas/```

Then restart gnome-shell, done. I'm happy that I can increase the maximum value anyway, because 200 is a little bit small for me.

Issue #109, opened 2018-01-24
FR: Directly paste from list


pat-richter:
I would like to not only set an item as "active" (and push it to my clipboard), but paste it directly in the currently focused field.

An example use case is:
[Shortcut] to open the list
[Arrow down]
[Arrow down]
[Arrow down]
[Shortcut] to paste the current entry (or maybe just "Enter")

This would remove the additional step of Ctr+V the value.

monkeydri:
@Tudmotu I will try to implement it, at least for Xorg it should not be this hard.

Tudmotu:
Sounds from that thread that it's not really possible, right?
In any case, feel free to give it a try! I would try to do this on Wayland tho, since Xorg is going to disappear soon. Maybe Wayland has an option to fake ctrl+v calls somehow. I must admit I have no clue how to do that.. :sweat_smile:

komakino:
@Tudmotu ClipIt uses xdotool so as long as you can run shell commands, it should be no problem with that tool. `/bin/sh -c 'xdotool key ctrl+v'`

daf121:
@komakino @Tudmotu Any luck using xdotool ? Anyone tried? :)

daveflr:
Yeah, I need this feature too, any advance of this?

gpothier:
Hi, I created a fork of this repo (https://github.com/gpothier/gnome-shell-extension-clipboard-indicator) and adapted @tatums pull request to the current version of the extension (and replaced `xdotool type` by `xdotool key` to simulate a ctrl+v press). This does not address the issues raised by @Tudmotu in the PR (namely, it only works in X11 and requires xdotool), so I don't know if it is worthwhile to create a new PR. However, it works for me.

PrOF-kk:
Looks like there's an xdotool "clone" that works independently of the window manager: [ydotool](https://github.com/ReimuNotMoe/ydotool), but there might be issues with licensing as it's under AGPLv3

SUPERCILEX:
Hey folks, this is available in a re-written version of the extension: https://github.com/SUPERCILEX/gnome-clipboard-history.

PR #168, opened 2019-10-01
adds paste on select

tatums:
This is an attempt to close https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/109 I've also included a config option "paste on select" (disabled by default)

I opted to use `xdotool type` interface.  I found this to be more reliable since, for example, gnome terminal uses `ctrl+shift+v` instead of `ctrl+v`.  

Screenshot%20from%202019-10-01%2008-09-37 png


Tudmotu:
Hi Tatums!
Thank you for taking the time to tackle this feature request :)
But there are a couple of issues:
1. This doesn't seem to work in Wayland. Maybe this article can point in the right direction?
https://fedoramagazine.org/simulate-device-input-evemu/
This has to support Wayland.
1. Both these programs (xdotool, evemu) are not installed by default - people would have to manually install these. What can be done about it?
1. Regarding using `type` api - does it work with unicode and such?

gpothier:
I applied these changes locally, but replacing _xdotool type_ by _xdotool key ctrl+v_, because although it should work in more cases, for some reason the type operation is **very** slow (several seconds). It works reasonably well with, though, except for instance in Terminal, or IntelliJ IDEA.

PR #179, opened 2019-10-11
Update clipboard-indicator.po

Etamuk:
Also the clipboard-indicator.mo file as zip (cannot upload it here directly):
[clipboard-indicator.zip](https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/files/3717336/clipboard-indicator.zip)

### DIFF
--- a/locale/de/LC_MESSAGES/clipboard-indicator.po
+++ b/locale/de/LC_MESSAGES/clipboard-indicator.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2019-10-11 13:17+0300\n"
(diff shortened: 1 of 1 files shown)


Tudmotu:
To update the .mo file, run the `compile-locales.sh` script, and commit the change.

Tudmotu:
I merged and updated the .mo files. Thanks !

Etamuk:
Thank you. I don't know much about programming. But I am good in translating ENG > DE. :-) 
When translating I always take the source file with the strings directly from the project page. 
I can not downloading the source and uploading it back to the branch again. I do not know enough about it. So sorry for that. For the most translation I am working on the .mo file is not needed.

Tudmotu:
That's ok, translating is already a great help - thanks! :)

PR #201, opened 2020-03-26
Allow installation for 3.36

mart-e:
_This PR only update the manifest, not tested if needs specific adaptations_

GNOME Shell 3.36 was released but it wasn't possible to install this application on it

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -5,7 +5,8 @@
         "3.28",
         "3.30",
         "3.32",
-        "3.34"
+        "3.34",
+        "3.36"
     ],
     "uuid": "clipboard-indicator@tudmotu.com",
     "name": "Clipboard Indicator",


jonafato:
As a testing anecdote, I ran this on 3.36 for several weeks with no issues.

rmader:
We should probably wait with this until https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/202 is resolved. Apart from that I also don't see any issues (apart from some warnings in the log that should get fixed sooner or later, but nothing critical)

Tudmotu:
Thanks!
This actually has no real effect I believe since 3.32. Extension can still be installed, just that 3.36 is "not supported".
I will merge this anyway :)

mart-e:
Screenshot from 2020-05-04 08-04-04

Really? I thought it was used for this list. Any idea how to add the 3.36 version there?
Anyway thanks for the merge :)

Tudmotu:
It does affect this dropdown, but you should still be able to install the "3.34" version on GNOME 3.36.

Issue #202, opened 2020-04-01
Not able to open settings from menu in 3.36.1


rmader:
This is due to a upstream change around the new extension app. We plan to add a new convenience function for that (it's already possible but a bit cumbersome).

Edit: see https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1163

memeplex:
I'm seeing this too, and it's also happening in other extensions.

memeplex:
Thanks!

PR #203, opened 2020-04-03
Use new convenience function to open settings

rmader:
GS 3.36 introduced a portal-like API for actions like opening settings.
That was done in order to make it possible to ship the new extensions
app in a sandboxed manner, e.g. via flatpak.

While talking to the API directly is a bit cumbersome, luckily there
is a convenience function we can make use of (from GS 3.36.2 on).

Closes https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/202

This depends on https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1163 to land first -> ~WIP~ -> MR landed now

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -553,10 +553,14 @@ const ClipboardIndicator = Lang.Class({
     },
 
     _openSettings: function () {
-        Util.spawn([
-            "gnome-shell-extension-prefs",
-            Me.uuid
-        ]);
+        if (typeof ExtensionUtils.openPrefs === 'function') {
+            ExtensionUtils.openPrefs();
+        } else {
+            Util.spawn([
+                "gnome-shell-extension-prefs",
+                Me.uuid
+            ]);
+        }
     },
 
     _initNotifSource: function () {


Tudmotu:
Thanks :)
Does this change not require anything more? All the listeners on the settings will work as expected?

rmader:
> Does this change not require anything more? All the listeners on the settings will work as expected?

That's how I understood Florian. AFAIK he didn't plan to break anything at all - this was just overlooked. And from a bit of testing everything appears to work fine (minus some bugs, but I they can be looked at independently)

fmuellner:
> Does this change not require anything more? All the listeners on the settings will work as expected?

The preference dialog was always opened from a separate process, so there shouldn't be a change in that regard.

Unless you do something very nasty like calling `Gio.Application.get_default()` to poke around in gnome-shell-extension-prefs itself, everything should be fine.

rmader:
Fixed the indentation. As the upstream change landed this should be good to go (apart from maybe making it more elegant - suggestions welcome :) )

rmader:
@Tudmotu: 3.36.2 was just released - can we go on with this one (and subsequently https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/201)? :)

---

It should be noted that [another change](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1220) in 3.36.2 that will allow the old method to work as long as the extension app is installed - so user might see #202 fixed even without this change.

Quote:
> While extensions are encouraged to switch to the new openPrefs()
convenience method added in commit 8030d9ad, restore the old
behavior with a small script under the old gnome-shell-extension-prefs
name that either calls out to the portal or launches the app.

So we should definitely land this :)

fmuellner:
> It should be noted that [another change](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1220) in 3.36.2 that will allow the old method to work as long as the extension app is installed

Two corrections:
 - when passing a valid UUID to the script, it doesn't matter whether the extensions app is installed or not
 - the change was clearly *meant* for 3.36.2, but as I'm an idiot I forgot to cherry-pick it until after the release

Issue #192, opened 2019-12-03
object.actor is deprecated


prasannavl:
```
Dec 03 11:42:44 pvl-x1 gnome-shell[2737]: Usage of object.actor is deprecated for ClipboardIndicator                                                                                                               
                                          get@resource:///org/gnome/shell/ui/environment.js:249:29                                                                                                                 
                                          _init@/home/pvl/.local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com/extension.js:88:9                                                                    
                                          wrapper@resource:///org/gnome/gjs/modules/_legacy.js:82:22                                                                                                               
                                          enable@/home/pvl/.local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com/extension.js:856:26                                                                 
                                          _callExtensionEnable@resource:///org/gnome/shell/ui/extensionSystem.js:132:13                                                                                            
                                          _enableAllExtensions/<@resource:///org/gnome/shell/ui/extensionSystem.js:495:17                                                                                          
                                          _enableAllExtensions@resource:///org/gnome/shell/ui/extensionSystem.js:494:13                                                                                            
                                          _sessionUpdated@resource:///org/gnome/shell/ui/extensionSystem.js:522:13                                                                                                 
                                          _emit@resource:///org/gnome/gjs/modules/signals.js:135:27                                                                                                                
                                          _sync@resource:///org/gnome/shell/ui/sessionMode.js:207:9                                                                                                                
                                          popMode@resource:///org/gnome/shell/ui/sessionMode.js:176:9                                                                                                              
                                          _continueDeactivate@resource:///org/gnome/shell/ui/screenShield.js:1203:13                                                                                               
                                          deactivate/<@resource:///org/gnome/shell/ui/screenShield.js:1189:39                                                                                                      
                                          finish@resource:///org/gnome/shell/gdm/authPrompt.js:492:13                                                                                                              
                                          finish@resource:///org/gnome/shell/ui/unlockDialog.js:134:9                                                                                                              
                                          deactivate@resource:///org/gnome/shell/ui/screenShield.js:1189:13                                                                                                        
                                          ScreenShield</</<@resource:///org/gnome/shell/ui/screenShield.js:536:52                                                                                                  
                                          _emit@resource:///org/gnome/gjs/modules/signals.js:135:27                                                                                                                
[...]

PR #207, opened 2020-04-16
Adding makefile

b00f:
It has lots of changes, please don't panic. It is for good of the project. What I have done:
I added a makefile to pack, install and test package. Packing and installing is using `gnome-shell` which is more simpler and easier. It compiles all po files and put them inside locale. 

Another good point of having makefile is you can clone your project outside the gnome-shell folder. 

Also I removed all mo files (which are not necessary) and re-organized the locale folder.

Regards,


Tudmotu:
Hi @b00f , thank you for the effort you put into this.
As we discussed in our emails I unfortunately can't merge this PR, but feel free to take on some issue from the "issues" page. I believe you fixed (or at least partially) #192, maybe you can open a PR just with that fix?
Thanks :)

Issue #206, opened 2020-04-16
Lagging and losing clipboard data 


b00f:
I have been using this extension for a while and I have seen sometimes when the extension is active I lost some clipboard data.
Imaging I cut a line, then for a second the os lags and instead of cutting one line, it cuts several lines. Also the clipboard-indicator shows only the last line. 

To find out what exactly is happening I put a log inside the `_processClipboardContent` function to print the clipboard text : `log("Clipboard text: "+ text);`
The I opened the file with vscode and try to cut lines by pressing `ctrl+x`. Here is the log result:
```
Apr 16 16:25:43 archlinux gnome-shell[29594]: Clipboard text: 43
Apr 16 16:25:43 archlinux gnome-shell[29594]: Clipboard text: 44
Apr 16 16:25:44 archlinux gnome-shell[29594]: Clipboard text: 45
Apr 16 16:25:44 archlinux gnome-shell[29594]: Clipboard text: 46
Apr 16 16:25:45 archlinux gnome-shell[29594]: Clipboard text: 47
Apr 16 16:25:45 archlinux gnome-shell[29594]: Clipboard text: 48
Apr 16 16:25:45 archlinux gnome-shell[29637]: Key repeat discarded, Wayland compositor doesn't seem to be processing events fast enough!
Apr 16 16:25:45 archlinux gnome-shell[29637]: Key repeat discarded, Wayland compositor doesn't seem to be processing events fast enough!
Apr 16 16:25:45 archlinux gnome-shell[29594]: Window manager warning: last_user_time (8013269) is greater than comparison timestamp (8012851).  This most likely represents a buggy client sending inaccurate timestamps in messages such as _NET_ACTIVE_WINDOW.  Trying to work around...
Apr 16 16:25:45 archlinux gnome-shell[29594]: Window manager warning: 0x800001 appears to be one of the offending windows with a timestamp of 8013269.  Working around...
Apr 16 16:25:45 archlinux gnome-shell[29594]: Clipboard text: 51
Apr 16 16:25:45 archlinux gnome-shell[29637]: Key repeat discarded, Wayland compositor doesn't seem to be processing events fast enough!
Apr 16 16:25:45 archlinux gnome-shell[29637]: Key repeat discarded, Wayland compositor doesn't seem to be processing events fast enough!
Apr 16 16:25:45 archlinux gnome-shell[29637]: Key repeat discarded, Wayland compositor doesn't seem to be processing events fast enough!
[...]

Issue #200, opened 2020-03-23
Conflict whit Libreoffice copy and paste


maurrox:
After a long search on the web i solved a serious problem whit libreoffice paste.
Randomly the paste of a cell or a selection of cell it wasn't working properly.
After disabling your extension everything went back to normal.
I try enabling and the problem return.
Tested on Debian 9, libreoffice 6.3.5.2
I'm sorry because your extension worked very well

rmader:
Is this on Wayland? In that case it's likely a duplicate of https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/194. I'm afraid the fix (in gtk-3.24.14) is unlikely to get backported to debian 9 though :/

Tudmotu:
@rmader currently the extension uses the new-ish "selection api", which I know has issues with Xwayland. Do you think that might be the cause of this?

rmader:
> @rmader currently the extension uses the new-ish "selection api", which I know has (or had) issues with Xwayland. Do you think that might be the cause of this?

Uh, it's Debian stretch, not buster (what I first though). Well, given that we're talking about Gnome/GTK 3.22, which is not supported by this extension any more - I guess this is a won't fix.

Tudmotu:
Ah, I missed this as well. @maurrox are you using the 3.22 version of this extension?

Issue #195, opened 2020-02-05
Current clipboard is not detected when changed by another tool


Whisprin:
I'm using [autocutsel](https://www.nongnu.org/autocutsel/) to sync my PRIMARY and CLIPBOARD selection.

When I select text, I can paste it using Ctrl+V, so the sync works fine. However, the text that is pasted using Ctrl+V doesn't appear in the clipboard history of this extension.

The same behaviour can be examined when using `xclip`. When xclip is used to set text in the clipboard this extensions doesn't pick up the changes.

---

My system: 
Arch Linux Kernel 5.4.15
Gnome 3.34.2
wayland 1.17.0
xwayland 1.20.7
gtk3 3.24.13

Tudmotu:
Hmm, I would guess that since this extension is using the new Gnome clipboard, and those tools communicate with the X11 clipboard, they don't align. I'm not sure if it can work, unfortunately :\

PR #215, opened 2020-05-05
Add new settings for clipboard listener

b00f:
Add new settings for Clipboard Listener. It can set to `Auto-detect` or `Timer`.
It addresses #206, However it doesn't fully fix it.


extension.js
```diff
@@ -557,7 +559,8 @@ const ClipboardIndicator = Lang.Class({
             ExtensionUtils.openPrefs();
         } else {
             Util.spawn([
-                "gnome-shell-extension-prefs",
+                "gnome-extensions",
+                "prefs",
```
RaphaelRochet, line 563: This has nothing to do with clipboard listener

b00f: Without this you are not able to ope the setting dialog. For testing the change it is necessary.

RaphaelRochet: Yep. But should not be in PR commit. You could had you 'lister' branch (re)based after that change. You other PR have this lines changed also : accepting one will cause conflict on the other.

b00f: I think this change is related to this PR.

prefs.js
```diff
@@ -91,9 +92,16 @@ const App = new Lang.Class({
         this.field_display_mode = new Gtk.ComboBox({
             model: this._create_display_mode_options()});
 
-        let rendererText = new Gtk.CellRendererText();
-        this.field_display_mode.pack_start (rendererText, false);
-        this.field_display_mode.add_attribute (rendererText, "text", 0);
+        let rendererDsiplayModeText = new Gtk.CellRendererText();
+        this.field_display_mode.pack_start (rendererDsiplayModeText, false);
+        this.field_display_mode.add_attribute (rendererDsiplayModeText, "text", 0);
```
RaphaelRochet, line 97: Dsiplay ? Your fingers seems to have their own spelling :wink: : 
```suggestion
        let rendererDisplayModeText = new Gtk.CellRendererText();
        this.field_display_mode.pack_start (rendererDisplayModeText, false);
        this.field_display_mode.add_attribute (rendererDisplayModeText, "text", 0);
```

RaphaelRochet:
Okay .. so this in fact is a way to force timer mode even if "metaDisplay.get_selection" is not available.
The combo box may induce error on being able to use 'auto-detect'.
Why not use a "Force timer mode (may fix lag issues)" switch ?

b00f:
>Why not use a "Force timer mode (may fix lag issues)" switch ?

I purposely chose combo box. Currently both methods are not perfect. We might find better API or library to track the Clipboard in future.

extension.js
Tudmotu: Why did you make this change?

Tudmotu:
@b00f Like we discussed, I think this should be a toggle that is labelled "Use legacy X11 clipboard".
It is:
1. More accurate
2. More obvious that it's related to issues with X11 apps.

b00f:
"legacy X11 clipboard" is a separate API, right?

Tudmotu:
Ok, after looking over the actual code and your changes, you convinced me that the dropdown options are OK. I would say I'm wrong in saying that it's "use x11 clipboard". It's more like "Avoid new GNOME API". But that makes it more confusing to users, so "Auto-Detect" / Timer is better.
But they still feel not ideal. How do they convey the difference, and the problem that it solves?
A user reading the label should be like "ah, that's what I was looking for", and I think these labels don't do that.
Do we have any more ideas for the labels?

b00f:
> Do we have any more ideas for the labels?

How about  having 3 options:
- Auto detect
- Gnome API 
- Using timer

b00f:
Another question is which one should be selected by default. I think timer can set as default value. 

These issues might be related to this PR: #206 #200 #195 
We can make timer as a default option, then monitor the issues. If users doesn't report related issues, we might consider closing theses issues.

b00f:
It's really bad that my PR is still here. If I knew you don't appreciate Open Source culture, I would't work on this application.

b00f:
May I know why you close this PR?

Tudmotu:
@b00f is this still happening to you?

b00f:
I have change my Linux to PopOS!
That was on time I had Arch Linux.
I don't think they have fixed it. The original bug is still open: https://gitlab.gnome.org/GNOME/mutter/-/issues/941

PR #216, opened 2020-05-05
ScrollView clip to allocation

b00f:
Clip the scroll view to the content.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -150,8 +150,9 @@ const ClipboardIndicator = Lang.Class({
 
             that.scrollViewFavoritesMenuSection = new PopupMenu.PopupMenuSection();
             let favoritesScrollView = new St.ScrollView({
-                style_class: 'ci-history-menu-section',
-                overlay_scrollbars: true
+                style_class: 'vfade ci-history-menu-section',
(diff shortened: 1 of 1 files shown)


RaphaelRochet:
I don't see any effect of `clip_to_allocation: true`, what is it for ?

b00f:
It clips the content of scroll-view to the allocated area. You might scroll down and up, then click the top-bar icon. Sometimes the scroll view pick wrong item which is not inside the allocated area.

RaphaelRochet:
Oh ok thanks for the explanation

b00f:
I searched a lot to find the solution. That is really bad of gnome developers. A simple thing but big effect.

b00f:
May I know why you close this PR?

Tudmotu:
What does it fix exactly? Can you share steps to reproduce? A screenshot?
Also, is there an open issue about it?

b00f:
I think comments on this PR are clear enough:

> Oh ok thanks for the explanation

> I searched a lot to find the solution. That is really bad of gnome developers. A simple thing but big effect.

Forget it. Let it be!

Issue #182, opened 2019-10-17
[Feature Request]About Clear History


nlpsuge:
Hi,

1. Please consider to add an option to remove Clear History from the main menu. I can sometimes misclick the Clear History, so I loss the whole history.

2. Please also consider to add a prompt so that I will know I've clicked the Clear History, then I decide how to handle it, clicking yes or no.

PR #217, opened 2020-05-05
Adding confirmation dialog for clear all the items

b00f:
Add confirmation dialog to the extension.

Addressing #182

### DIFF
--- a/confirmDialog.js
+++ b/confirmDialog.js
@@ -4,19 +4,19 @@ const ModalDialog = imports.ui.modalDialog;
 const CheckBox = imports.ui.checkBox;
 const Clutter = imports.gi.Clutter;
 
-const Gettext = imports.gettext.domain("gnome-trash");
+const Gettext = imports.gettext.domain("clipboard-indicator");
 const _ = Gettext.gettext;
 
 
 
-function openConfirmDialog(title, message, sub_message, ok_label, callback) {
-    new ConfirmDialog(title, message + "\n" + sub_message, ok_label, callback).open();
+function openConfirmDialog(title, message, sub_message, ok_label, cancel_label, callback) {
+    new ConfirmDialog(title, message + "\n" + sub_message, ok_label, cancel_label, callback).open();
 }
 
 const ConfirmDialog = GObject.registerClass(
   class ConfirmDialog extends ModalDialog.ModalDialog {
 
-    _init(title, desc, ok_label, callback) {
+    _init(title, desc, ok_label, cancel_label, callback) {
       super._init();
 
       let main_box = new St.BoxLayout({
@@ -46,7 +46,7 @@ const ConfirmDialog = GObject.registerClass(
 
       this.setButtons([
         {
-          label: _("Cancel"),
+          label: cancel_label,
           action: () => {
             this.close();
           },
--- a/confirmDialog.js
+++ b/confirmDialog.js
@@ -1,0 +1,65 @@
+const St = imports.gi.St;
+const GObject = imports.gi.GObject;
+const ModalDialog = imports.ui.modalDialog;
+const CheckBox = imports.ui.checkBox;
+const Clutter = imports.gi.Clutter;
+
+const Gettext = imports.gettext.domain("gnome-trash");
+const _ = Gettext.gettext;
+
+
+
+function openConfirmDialog(title, message, sub_message, ok_label, callback) {
+    new ConfirmDialog(title, message + "\n" + sub_message, ok_label, callback).open();
+}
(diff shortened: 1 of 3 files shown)


extension.js
```diff
@@ -332,22 +333,27 @@ const ClipboardIndicator = Lang.Class({
...
-            if (!mItem.currentlySelected) {
-                let idx = that.clipItemsRadioGroup.indexOf(mItem);
-                mItem.destroy();
-                that.clipItemsRadioGroup.splice(idx,1);
-            }
+        const title = _("Clear all?");
+        const message = _("Are you sure you want to delete all clipboard items?");
+        const sub_message = _("This operation cannot be undone.");
+
+        ConfirmDialog.openConfirmDialog(title, message, sub_message, _("Clear"), () => {;
```
RaphaelRochet, line 341: I wonder .. why did you use consts for all strings but ok_label ?

confirmDialog.js
```diff
@@ -0,0 +1,65 @@
+const St = imports.gi.St;
+const GObject = imports.gi.GObject;
+const ModalDialog = imports.ui.modalDialog;
+const CheckBox = imports.ui.checkBox;
+const Clutter = imports.gi.Clutter;
+
+const Gettext = imports.gettext.domain("gnome-trash");
```
RaphaelRochet, line 7: You forgot to edit this ? We should use extension context

RaphaelRochet:
A shame that GS does not provide a simple confirm() function ...
A minimum of tweaking could make this dialog perfect !
Here is a side-by-side with a gedit confirm dialog to help :
image

confirmDialog.js
b00f: Good find. Copy-pasting issue. Thanks man!

b00f:
> A minimum of tweaking could make this dialog perfect !

Can you explain what do you mean? You mean make it wider?

RaphaelRochet:
No only tweak font size and spacings. Very minor things ... :smile: 
Do you think there's a "right" way to use color on "Clear" button ? I mean without breaking themes ...
Like this dialog in nautilus : 
image

extension.js
b00f: Done! As requested.

confirmDialog.js
```diff
@@ -4,19 +4,19 @@ const ModalDialog = imports.ui.modalDialog;
 const CheckBox = imports.ui.checkBox;
 const Clutter = imports.gi.Clutter;
 
-const Gettext = imports.gettext.domain("gnome-trash");
+const Gettext = imports.gettext.domain("clipboard-indicator");
 const _ = Gettext.gettext;
```
RaphaelRochet, line 9: You can remove these now. This will make this confirm-dialog class reusable without a change :+1:

b00f: Unfortunately the translation won't work if I remove it. I don't like it as well.

b00f:
We can modify CSS file, But I prefer to not. Keep it as it is. Gnome is different from distro to distro.

confirmDialog.js
```diff
@@ -0,0 +1,65 @@
...
+
+      let subject_label = new St.Label({
+        style: `font-weight: 700`,
+        text: title
+      });
+
+      message_box.add(subject_label, { y_fill: true, y_align: St.Align.START });
+
+      let desc_label = new St.Label({
+        style: 'padding-top: 10px; padding-bottom: 20px;',
```
RaphaelRochet, line 41: Can't argue with theming, but at least remove double bottom space, and follow gnome guidelines on element spacing :
```suggestion
        style: 'padding-top: 12px; ',
```

confirmDialog.js
RaphaelRochet: I just update the French translation with these strings, and removed these 2 lines : works like a charm :)

b00f: How about `Gettext.gettext`? Do you import gettext?

RaphaelRochet: I removed both lines. There's no use for "_" in that file.

confirmDialog.js
```diff
@@ -0,0 +1,65 @@
+const St = imports.gi.St;
+const GObject = imports.gi.GObject;
+const ModalDialog = imports.ui.modalDialog;
+const CheckBox = imports.ui.checkBox;
+const Clutter = imports.gi.Clutter;
+
+const Gettext = imports.gettext.domain("clipboard-indicator");
+const _ = Gettext.gettext;
```
RaphaelRochet, line 8: Like said.
```suggestion
```

PR #232, opened 2020-06-22
Update clipboard-indicator.po

albanobattistella:


Tudmotu:
Thank you so much for taking the time to create this PR. Apologies for the late reply. There was an earlier PR that already added italian translations, so I'm closing this one.

PR #234, opened 2020-08-15
Add simple Makefile

rmader:
Move the script files into a Makefile and add a `install` command,
making development somewhat easier depending on the workflow.
As the list of files is needed for both installing and bundling,
make both use a shared list.


Tudmotu:
@rmader I don't understand what is the "install" necessary for?
Also, if you don't mind I would appreciate if you can:
1. Update the "modules" to include the new "confirmDialog.js" file
2. Also add the "update-po-files" and "compile-locales" to the makefile? Then we can remove them as well.

Thanks!

rmader:
> @rmader I don't understand what is the "install" necessary for?

I personally find it very useful to have the git repo independent of the installed extension folder, so having a install command makes it easy to install exactly the version users would get from a bundle during development.

> Also, if you don't mind I would appreciate if you can:
> 
>     1. Update the "modules" to include the new "confirmDialog.js" file
Done
>     2. Also add the "update-po-files" and "compile-locales" to the makefile? Then we can remove them as well.
Done
> Thanks!

This is now hopefully good to go now.

Tudmotu:
Very late, but thank you for this PR! :)

PR #273, opened 2021-02-28
update topbar preview on selection from menu

### LINKED ISSUE #210: Selecting an old clipping from the list does not update the preview in the top bar
When I copy a new clipping using Control+C, the preview text in to top bar gets updated accordingly. 

Nevertheless, if I select an old clipping from the list, I can paste it, but the preview text in top bar does not reflect the change.

vaibhvshrma:
Hey @Tudmotu, thanks for the amazing tool, I do not know how I can work without it! 
This fixes #210, and updates the preview in the following cases which were not there earlier:

- When the extension is restarted
- Previous or next content selected through keyboard bindings
- Another selection is made through the dropdown menu

Tested locally. Let me know if you need anything else.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -409,6 +409,9 @@ const ClipboardIndicator = Lang.Class({
     _selectMenuItem: function (menuItem, autoSet) {
         let fn = Lang.bind(menuItem, this._onMenuItemSelected);
         fn(autoSet);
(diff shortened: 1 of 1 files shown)


kshitijzutshi:
I checked the tool in Gnome 3.38.3 version. Seems like its working.
@vaibhvshrma I am new to this open source contributing thing. I like the clipboard tool too. Could you help me get up to speed with the project structure of this tool? Perhaps connect? Please revert back.
Thanks :)

PR #277, opened 2021-03-13
Add support for GNOME 40

frandieguez:
Hi,

please consider merging these changes. I have updated the prefs panel to support GNOME 40

Regards

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,6 +1,7 @@
 {
     "shell-version": [
-        "3.38"
+        "3.38",
+        "40"
     ],
     "uuid": "clipboard-indicator@tudmotu.com",
     "name": "Clipboard Indicator",
--- a/prefs.js
+++ b/prefs.js
@@ -48,7 +48,10 @@ const App = new Lang.Class({
     Name: 'ClipboardIndicator.App',
     _init: function() {
         this.main = new Gtk.Grid({
-            margin: 10,
+            margin_top: 10,
+            margin_bottom: 10,
+            margin_start: 10,
+            margin_end: 10,
             row_spacing: 12,
             column_spacing: 18,
             column_homogeneous: false,
@@ -183,8 +186,8 @@ const App = new Lang.Class({
                 let inputWidget = input;
 
                 if (input instanceof Gtk.Switch) {
-                    inputWidget = new Gtk.HBox();
-                    inputWidget.pack_end(input, false, false, 0);
+                    inputWidget = new Gtk.Box();
+                    inputWidget.append(input, false, false, 0);
                 }
 
                 if (label) {
@@ -226,7 +229,8 @@ const App = new Lang.Class({
         SettingsSchema.bind(Fields.STRIP_TEXT, this.field_strip_text, 'active', Gio.SettingsBindFlags.DEFAULT);
         SettingsSchema.bind(Fields.ENABLE_KEYBINDING, this.field_keybinding_activation, 'active', Gio.SettingsBindFlags.DEFAULT);
 
-        this.main.show_all();
+        if (typeof this.main.show_all === 'function')
+            this.main.show_all();
     },
     _create_display_mode_options : function(){
         let options = [{ name: _("Icon") },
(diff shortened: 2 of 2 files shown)


prefs.js
```diff
@@ -226,7 +229,8 @@ const App = new Lang.Class({
         SettingsSchema.bind(Fields.STRIP_TEXT, this.field_strip_text, 'active', Gio.SettingsBindFlags.DEFAULT);
         SettingsSchema.bind(Fields.ENABLE_KEYBINDING, this.field_keybinding_activation, 'active', Gio.SettingsBindFlags.DEFAULT);
 
-        this.main.show_all();
+        if (this.main.show_all)
```
rmader, line 232: Maybe `typeof this.main.show_all === 'function'`?

metadata.json
```diff
@@ -1,8 +1,7 @@
 {
     "shell-version": [
         "3.38",
-        "40.beta",
-        "40.rc"
+        "40.0"
```
rmader, line 4: Maybe just use `40` and squash the commit? That's what's done for the officially supported ones at https://gitlab.gnome.org/GNOME/gnome-shell-extensions/

Makefile
```diff
@@ -0,0 +1,27 @@
+PREFIX ?= ~/.local/share/gnome-shell/extensions
+
+TRANSLATION_ID=clipboard-indicator
+EXTENSION_UUID=clipboard-indicator@tudmotu.com
+
+all: build install
```
rmader, line 6: I'd drop the `install`, assuming most people do not expect a `install` when just running `make`. Also, maybe move this to a own commit?

frandieguez:
Hi! I'll fix all the issues you've find and fix them.

rmader:
> Hi! I'll fix all the issues you've find.

Thanks! Concerning the Makefile: maybe we want to go with https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/234 instead as it does a bit more than the one here.

frandieguez:
> > Hi! I'll fix all the issues you've find.
> 
> Thanks! Concerning the Makefile: maybe we want to go with #234 instead as it does a bit more than the one here.

yes, I'll revert that change. I like the way it works the other Makefile, as it goes all the way porting from .sh files to final Makefile, which is what I'll like to do in the first place.

metadata.json
rmader: FTR, I'd recommend using really just `40` instead of `40.0`

rmader: To explain: from 40 on the major version signals API stability - independent of the point release or beta/release candidate. And API stability is everything we are interested in as extension.

frandieguez: Uploading new version, right now.

frandieguez (commented):
I think that it's all fixed. Please tell me if you need more changes

prateekmedia:
Please also get rid of dropdown button as it looks ugly with gnome 40 design and also because they removed it from the power and network strip.

prateekmedia:
Also one thing to notice is that the keyboard shortcuts inside the preferences says undefined but we can still set it, can you look into it please.

prefs.js
```diff
@@ -183,8 +186,8 @@ const App = new Lang.Class({
                 let inputWidget = input;
 
                 if (input instanceof Gtk.Switch) {
-                    inputWidget = new Gtk.HBox();
-                    inputWidget.pack_end(input, false, false, 0);
+                    inputWidget = new Gtk.Box();
+                    inputWidget.append(input);
```
prateekmedia, line 190: Append would not work in 3.38, either you should check first that is it valid, so that you could use pack_start for 3.38, or else you should remove 3.38 support.

prateekmedia (changes requested):

prefs.js
frandieguez: Should be fixed

frandieguez:
Please check https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/280 that PR is more complete

Tudmotu:
@frandieguez Thank you for taking the time to send this PR, I really appreciate it!
But I'm gonna go with #280 so I am closing this one. But really - thank you for opening this PR as well :pray:

PR #280, opened 2021-03-25
Gnome Shell 40 Port

justperfection:
**extension.js**

Removed the deprecated actor for ClipboardIndicator.

**prefs.js**

- _Gtk.accelerator_parse()_ now returns first element as ok.
- _show_all()_ no longer works and needed on GTK4.
- Instead of _HBox_ use Gtk.Box.
- _margin_ no longer supported in GTK4 and replaced with more detailed margin properties.

**metadata.json**

Removed "3.38" and added "40.0" to the shell version.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,6 +1,6 @@
 {
     "shell-version": [
-        "40.0"
+        "40"
     ],
     "uuid": "clipboard-indicator@tudmotu.com",
     "name": "Clipboard Indicator",
--- a/metadata.json
+++ b/metadata.json
@@ -1,6 +1,6 @@
 {
     "shell-version": [
-        "3.38"
+        "40.0"
     ],
     "uuid": "clipboard-indicator@tudmotu.com",
     "name": "Clipboard Indicator",
--- a/extension.js
+++ b/extension.js
@@ -90,7 +90,7 @@ const ClipboardIndicator = Lang.Class({
         hbox.add_child(this._buttonText);
         this._downArrow = PopupMenu.arrowIcon(St.Side.BOTTOM);
         hbox.add(this._downArrow);
-        this.actor.add_child(hbox);
+        this.add_child(hbox);
 
         this._createHistoryLabel();
         this._loadSettings();
--- a/prefs.js
+++ b/prefs.js
@@ -48,7 +48,10 @@ const App = new Lang.Class({
     Name: 'ClipboardIndicator.App',
     _init: function() {
         this.main = new Gtk.Grid({
-            margin: 10,
+            margin_top: 10,
+            margin_bottom: 10,
+            margin_start: 10,
+            margin_end: 10,
             row_spacing: 12,
             column_spacing: 18,
             column_homogeneous: false,
@@ -183,8 +186,8 @@ const App = new Lang.Class({
                 let inputWidget = input;
 
                 if (input instanceof Gtk.Switch) {
-                    inputWidget = new Gtk.HBox();
-                    inputWidget.pack_end(input, false, false, 0);
+                    inputWidget = new Gtk.Box({orientation: Gtk.Orientation.HORIZONTAL,});
+                    inputWidget.append(input);
                 }
 
                 if (label) {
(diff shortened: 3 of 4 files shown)


frandieguez:
Hope we get this PR merged. I have a similar PR from some days ago but I don't know if it was enough to be merged... https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/277

justperfection:
@frandieguez Sorry, didn't see that one. I'm just going over all popular extensions and porting them.

btw, we don't need `show_all()` and we don't need to check that with `typeof`. dropping that is safe for GNOME Shell +3.36.

Anyway, we shouldn't add 3.36 and 3.38 support for this one since I've mentioned about `Gtk.accelerator_parse()`. I've added this one to the port guide too and soon will be public.

If supporting older versions is the end, I can add version checking.

prateekmedia:
Please also get rid of dropdown button as it looks ugly with gnome 40 design and also because they removed it from the power and network strip.

justperfection:
@prateekmedia the prefs has dropdown visibility option and works on GNOME Shell 40.

metadata.json
```diff
@@ -1,6 +1,6 @@
 {
     "shell-version": [
-        "3.38"
+        "40.0"
```
rmader, line 3: Note that `40.0` in the new version scheme is equivalent to `3.38.0` in the old one, i.e. this will stop working as soon as `40.1` is released in a couple of weeks.

justperfection: `40` is now a supported version on e.g.o website.

prateekmedia:
@justperfection I don't know enough, but I had done the same i.e. 40, 41, 42 for my extension without .0 , but it is showing incompatible in EGO with gnome 40. Here's the link 
https://extensions.gnome.org/extension/3724/net-speed-simplified/

justperfection:
@prateekmedia "40" is the accepted version since yesterday. This is only applies to the new submitted extensions and not pending for review extensions.

rmader:
> @prateekmedia the prefs has dropdown visibility option and works on GNOME Shell 40.

@justperfection could you change the default behaviour to disabled, in order to match default 40? IMO it would be even cleaner to just drop the future, but that can probably wait a bit more.

justperfection:
@rmader done!

RuiGuilherme:
I made a feature to save images in Clipboard History(#283 ) and I just tested with this patch and it worked perfectly.

abhaypatil2000:
@justperfection it doesn't work for arch linux. I git cloned [your repo](https://github.com/justperfection/gnome-shell-extension-clipboard-indicator/tree/gnome-shell-40) in extension folder and then restarted the gnome shell. And then opened extensions app and still it shows not compatible with your gnome version. I am using arch linux with gnome 40

justperfection:
@abhaypatil2000 because are not using my gnome-shell-40 branch.

RuiGuilherme:
Works on Arch;
Captura de tela de 2021-04-09 03-30-52

Btw, it works fine with my patch for images support. #283

RuiGuilherme:
Works on Arch;

Captura de tela de 2021-04-09 03-30-52

Btw, it works fine with my patch for images support. #283

verbalius:
tested on Arch and it works without problems. Keyboard shortcuts, all settings and notifications. 
image

rmader:
@Tudmotu would love to see you finding time to land this :)

In case you are busy with other stuff, maybe we could have additional maintainers (I'd be open to volunteer for that, probably others as well) or even move the extension to the officially supported ones (https://gitlab.gnome.org/GNOME/gnome-shell-extensions/) in order to keep it working on new GS versions.

Tudmotu:
Thank you @justperfection for this PR! :clap: :smiley: 
This is very appreciated. Thank you. :pray:

Tudmotu:
Uploaded to e.g.o, will hopefully be approved in the next couple of days.

Tudmotu:
New version is now approved on e.g.o :)
Thanks everyone for pushing this fix

PR #283, opened 2021-04-03
Save image clipboard history

RuiGuilherme:
I believe that [st_clipboard_get_content()](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/master/src/st/st-clipboard.c) have a bug that prevents the use of Image / PNG, because of this issue the images are being saved without the Alpha Channel. 

How test:
**checkout** and **pull**;;;; _**origin save_history_image**_ inside ~/.local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com 

I opened this issue in the Gnome Shell: https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/4034

### DIFF
--- a/byte64.js
+++ b/byte64.js
@@ -1,0 +1,133 @@
+/*
+MIT License
+Copyright (c) 2020 Egor Nepomnyaschih
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


RuiGuilherme:
Issue with **image/png** are solved: https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/4034#note_1074065

Now is just wait! When the Mutter receives the patch will only be necessary to replace **image/jpeg** to **image/png**.

PR #285, opened 2021-04-10
Added support for Gnome-shell 40

hmmanshh:
included gnome-shell 40 in metadata


prateekmedia:
There are already two PR's related to this, with more features rather than adding shell 40.0 in metadata.

Tudmotu:
Closing in favor of #280

Issue #222, opened 2020-05-09
Max History Size


lots0logs:
What is the purpose of limiting the max size a user can set? IMO, 200 is way too low. Why not let the user set whatever value is best for their workflow?

b00f:
Did you try with 200 items? Huge items inside the list might cause extension become unstable because of the performance. We need to make a benchmark on that.

lots0logs:
I actually changed the max value in my local copy so I could set it to 2000. So far so good but I'm guessing at some point it might start hindering performance?

b00f:
@Tudmotu Do you think maximum history can be unlimited? User might risk on performance.

lots0logs:
I think it should be up to the user since users will have various amounts of resources. Maybe add a disclaimer above or below the setting that states 

>Depending on your system's available resources, the value of this setting can cause performance issues if it is too high. In such cases you should try lowering the value until performance is acceptable.

pablodz:
How to change, I need 10k

Tudmotu:
10k will crash you system

pablodz:
Thanks for the recommendation

SUPERCILEX:
Hey folks, this is fixed in a re-written version of the extension: https://github.com/SUPERCILEX/gnome-clipboard-history.

pablodz:
Thanks mate

Issue #112, opened 2018-01-25
Support for images


orschiro:
This is such a brilliant extension!

Do you plan to add support for images?

Storing and pasting images from clipboard is quite essential for me and maybe others, too.

orschiro:
Posted a bounty for this. :-)

https://www.bountysource.com/issues/54210016-support-for-images

Tudmotu:
Hi @orschiro :)
Unfortunately lately I did not have time to work on this extension. Most of the latest features were implemented by community members.

If no one takes a look at it, I will probably get to it some when, but could take a while.

orschiro:
Hey @Tudmotu no problem! 

> Most of the latest features were implemented by community members.

Maybe someone from the community has an idea how to implement this one. :-)

Issue #21, opened 2015-06-29
Setting HISTORY_SIZE problem


jusa213:
Hello

Thanks for this useful extensions.
I tried to set the history size to 500 so I changed the values in prefs.js

> this.field_size = new Gtk.SpinButton({
>             adjustment: new Gtk.Adjustment({
>                 lower: 1,
>                 upper: 500,
>                 step_increment: 1

and in extensions.js to

> let MAX_ENTRY_LENGTH     = 500

So with GNOME Tweak Tool I can set the history size up to 500 but if I close the settings screen and reopen it, the history size is back to 50. How to change history size the right way?

Thx for helping out here.

Tudmotu:
Hi :)
I believe you need to edit the schema file (the ``.gschema.xml`` file in ``schemas/`` directory), then recompile the schema file using ``glib-compile-schemas``, and reload the shell.

*But* I highly recommend that you don't change it to more than 50, because this extension is not built to endure a large cache. It will probably slow down your entire shell. :\

PR #286, opened 2021-04-16
[WIP] Support for images

RuiGuilherme:
Features: 

- Change history size: **MAX** 2000 and **MIN** 2
- Support to save images **(Optional)**
- Image preview on menu https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/286/commits/d6475bf95cce05130b4460e3d920bcf08e038664
- Support for image/png **(Optional)**
- Support to file manager (Tested in **Dolphin**, **Nautilus** and **Nemo**) https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/commit/7a97da53d0c25aedaf503d1cf2a4cc3831a60abc

Know issue:

- Large images (9 mb or more) may not save. (8k wallpaper for example haha)
- image/png only works with Mutter 41.3+ ([patch mutter](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1812) )


RuiGuilherme:
Probably related:
#112 
#70 
#222  
#154 
#21

Tudmotu:
Hi :) 
Thank you for this, I hope you can make this work as this is a feature I personally have been waiting for.

I see you are serializing images as Base64 and save it inside the json, right?
What do you think about saving the images on the file system, and just reference their path in the JSON?

RuiGuilherme:
> I see you are serializing images as Base64 and save it inside the json, right?
Yes, but for _addEntry() I use bytes because for some reason convert Base64 to Bytes inside Clipboard.set_content() costs me ~2 Seconds for small imagens.

> What do you think about saving the images on the file system, and just reference their path in the JSON?
I will try using JSON, that should probably avoid the "throttling".

RuiGuilherme:
[commit bc45597](https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/286/commits/bc455972e9ca5f78711818db90be6b952ea62aef) now save files inside .cache and JSON save only path... 

I did several tests with and without the mutter patch... All works fine.

### Fix:
- 1mb images or more large can cause a "throttling".
- many images in history also cause "throttling".
- can cause delay on gnome start and restart


### Know issue:
- image/png works only with this [patch mutter](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1812)
- Large images (9 mb or more) may not save. (8k wallpaper for example haha)

Tudmotu:
Hi @RuiGuilherme :) 
Is this still WIP?

RuiGuilherme:
Yup, I need more testers... it works well on my i7 9750h but I don't have low-end spec to test possibles issues performance.

I will try to find someone available in some group related to GNOME to test.

utils.js
```diff
@@ -14,6 +14,40 @@ const REGISTRY_FILE = 'registry.txt';
...
+            const stream = obj.replace_finish(res);
+
+            stream.write_bytes_async(buffer, GLib.PRIORITY_DEFAULT, null, function (w_obj, w_res) {
+                w_obj.write_bytes_finish(w_res);
+                stream.close(null);
+            });
+        });
+    }
+
+    return path;
```
tcanabrava, line 38: since you are .create_async() the file here, are you sure that the file will exist when this function returns?

canadaduane:
I gave this a spin, but I think there is a bug when copying regular text:

1. With a fresh install/cache, copy some text
2. View the Clipboard Indicator history via clicking top bar icon (drop-down)
3. Copied text appears as a filepath, rather than the copied text itself

The filepath also appears to be incorrect: `/home/duane/.cache/clipboard-indicator@tudmotu.com/undefined`

extension.js
```diff
@@ -32,9 +33,13 @@ const Me = ExtensionUtils.getCurrentExtension();
 const Utils = Me.imports.utils;
 const ConfirmDialog = Me.imports.confirmDialog;
 const Prefs = Me.imports.prefs;
+
 const prettyPrint = Utils.prettyPrint;
 const writeRegistry = Utils.writeRegistry;
 const readRegistry = Utils.readRegistry;
+const writerImage = Utils.writerImage;
```
canadaduane, line 40: Should this be called `writeImage`?

extension.js
```diff
@@ -466,26 +514,71 @@ const ClipboardIndicator = Lang.Class({
...
 
         Clipboard.get_text(CLIPBOARD_TYPE, function (clipBoard, text) {
-            that._processClipboardContent(text);
+            if (text === null && SAVE_IMAGE) {
+                Clipboard.get_content(CLIPBOARD_TYPE, SUPPORT_PNG ? 'image/png' : 'image/jpeg', function (clipboard, content) {
+                    that._processClipboardContent(content.get_data(), 'image', content.hash());
+                })
+            }
+            else {
+                const itens = text.split('\n');
```
canadaduane, line 523: I think `text` can be null here. Consider:

1. let `text` be null
2. let `SAVE_IMAGE` be false

In that case, we fall through to the `else` block on line 522, and this line will fail when it tries to split a null value.

Also, as an aside: should this variable (`const itens`) be named `items`?

extension.js
```diff
@@ -466,26 +514,71 @@ const ClipboardIndicator = Lang.Class({
...
+        }
+        else {
+            content = writerImage(byteContent, name);
         }
 
-        if (text !== "" && text) {
+        if (content !== "" && content) {
             let registry = that.clipItemsRadioGroup.map(function (menuItem) {
-                return menuItem.clipContents;
+                return menuItem.clipContents
```
canadaduane, line 576: Leave semi-colon here?

extension.js
```diff
@@ -494,24 +587,24 @@ const ClipboardIndicator = Lang.Class({
...
-        this._addEntry(item.clipContents, item.clipFavorite, item.currentlySelected, false);
+        this._addEntry(item.clipContents, item.clipFavorite, item.currentlySelected, false, type ?? item.type);
     },
 
-    _findItem: function (text) {
-        return this.clipItemsRadioGroup.filter(
-            item => item.clipContents === text)[0];
+    _findItem: function (content, type) {
+            return this.clipItemsRadioGroup.filter(
+                item => item.clipContents === content)[0];
```
canadaduane, line 607: Formatting--looks like extra tabs were added here?

extension.js
```diff
@@ -682,6 +775,8 @@ const ClipboardIndicator = Lang.Class({
         TOPBAR_DISPLAY_MODE  = this._settings.get_int(Prefs.Fields.TOPBAR_DISPLAY_MODE_ID);
         DISABLE_DOWN_ARROW   = this._settings.get_boolean(Prefs.Fields.DISABLE_DOWN_ARROW);
         STRIP_TEXT           = this._settings.get_boolean(Prefs.Fields.STRIP_TEXT);
+        SAVE_IMAGE           = this._settings.get_boolean(Prefs.Fields.SAVE_IMAGE);
+        SUPPORT_PNG           = this._settings.get_boolean(Prefs.Fields.SUPPORT_PNG);
```
canadaduane, line 779: Line up equals signs?

extension.js
```diff
@@ -466,26 +514,71 @@ const ClipboardIndicator = Lang.Class({
...
 
-    _processClipboardContent (text) {
+    // type most be image or text
+    _processClipboardContent (byteContent, type, name) {
         const that = this;
-
-        if (STRIP_TEXT) {
-            text = text.trim();
+        let content;
+        if (STRIP_TEXT && type === "text") {
```
canadaduane, line 567: I think we're missing the case where STRIP_TEXT is false here. Shouldn't `content` be set to `byteContent` regardless?

canadaduane: e.g. ```        if (type === "text") {
            if (!STRIP_TEXT) {
                content = byteContent;
            }
            else {
                content = byteContent.trim();
            }
        }
        else {
            content = writerImage(byteContent, name);
        }
```

**Comment by canadaduane** [comment]
I've made a PR to @RuiGuilherme's branch https://github.com/RuiGuilherme/gnome-shell-extension-clipboard-indicator/pull/3

(My first time making a PR to a PR!)

**Comment by canadaduane** [comment]
FWIW, this works for me now. Thanks @RuiGuilherme for all your work!

**Comment by canadaduane** [comment]
@Tudmotu can this be merged in do you think?

**Comment by phamthainb** [comment]
This is so good, why is it not merged yet :clap:

**Comment by Tudmotu** [comment]
Hi all,
Am I missing something - when I right-click save image, it does not get added to the clipboard. Does this happen to you as well?

**Comment by Tudmotu** [comment]
Also, ctrl+v doesn't seem to paste images when I select them from the clipboard.

**Comment by Tudmotu** [comment]
I'm using GNOME 41 btw.

**Comment by RuiGuilherme** [comment]
> Hi all, Am I missing something - when I right-click save image, it does not get added to the clipboard. Does this happen to you as well?

I thought some modification of version 41 caused the bugs, so I made a new branch to test and everything works. 
https://youtu.be/r23P8MXqSx0

https://github.com/RuiGuilherme/gnome-shell-extension-clipboard-indicator/tree/Gnome41

**Comment by Tudmotu** [comment]
Ok, interesting. Thanks :)

**Comment by Tudmotu** [comment]
I made some small change due to conflicts and backwards compatibility, maybe I introduced a bug. Will look into it.

**Comment by hyuri** [comment]
Great features guys! I've been waiting for this!
Just wanted to chime in to ask if saving the images on disk could pose a potential privacy issue. In case, for example, that an outside agent gained access to the folder you're storing the images in.

**Comment by RuiGuilherme** [comment]
> Great features guys! I've been waiting for this! Just wanted to chime in to ask if saving the images on disk could pose a potential privacy issue. In case, for example, that an outside agent gained access to the folder you're storing the images in.

Yup, this can happen. However, this is not a specific problem of this feature, a malicious application could access JSON where the texts are stored and have access to all user history. 

But I believe this is not exactly something to worry about, since a malicious application could have access to the clipboard.

**Comment by ilazaridis** [comment]
I am also looking forward to this one. Image support would be really cool! :)

**Comment by mpetuska** [comment]
Just want to put this on the table - GPaste has supported images for a while now and it works quite well. It might be good to dig through their code for ideas?

File: prefs.js [hunk 8]
```diff
@@ -179,6 +183,17 @@ const App = new Lang.Class({
             hexpand: true,
             halign: Gtk.Align.START
         });
+        let saveImageLabel = new Gtk.Label({
+            label: _("Save images on history"),
+            hexpand: true,
+            halign: Gtk.Align.START
+        });
+        // need while this MR is pending: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1812
+        let pngImageLabel = new Gtk.Label({
```
SUPERCILEX, line 192: That PR was merged, is the workaround still necessary? (Assuming people update to the latest mutter.)

SUPERCILEX (commented):
I think this PR could benefit from a fundamentally different approach: instead of special casing images, why not support all file types? This would let us support random stuff like copied google slides, etc. If we don't understand the file type, we simply show its file path (maybe in *italics* to indicate that the path string is not what the user copied). If the path isn't available, we just show the mime type.

In a bit more detail, the idea looks like this:
- Accept any arbitrary clipboard contents
- Add `Save non-text clipboard items to disk` option (disabled by default)
  - If this option is true, we write the contents into our clipboard registry and cache dir
  - If not, we only keep the clipboard contents in memory so as to not waste disk space/cycles
- If the contents is a supported image, we display it in the clipboard UI

From a PR perspective, this offers some nice splits. First write a PR that supports arbitrary clipboard contents in memory. Then write a PR that displays known types (images). Then write a PR that adds the option to save stuff to disk.

SUPERCILEX:
I just tried this out and several things seem broken:
- My clipboard history was nuked, so backwards compat is not being handled properly
- The settings, private mode, and clear history buttons disappeared
- Copying an image from the web did not work (right click, copy image)
- Copying an image from the file manager *did* work, but the aspect ratio is broken and the image is too small
- After selecting a previously copied image, I am unable to paste it for whatever reason

Screenshot:
image

Tudmotu:
Thank you @SUPERCILEX for confirming the issues I encountered.
I opened a new PR that includes backward-compat fixes:
https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/334/files

To keep working on this, need to clone the `feature/cache-images` branch and to propose changes need to open a PR that points to `feature/cache-images` branch and not `master`.

RuiGuilherme:
> I just tried this out and several things seem broken:
> 
>     * My clipboard history was nuked, so backwards compat is not being handled properly
> 
>     * The settings, private mode, and clear history buttons disappeared
> 
>     * Copying an image from the web did not work (right click, copy image)
> 
>     * Copying an image from the file manager   _did  _ work, but the aspect ratio is broken and the image is too small
> 
>     * After selecting a previously copied image, I am unable to paste it for whatever reason
> 
> 
> Screenshot: image

Thank you. I'll work on the fixes. 
Did you use Branch for version 40+? 
https://github.com/RuiGuilherme/gnome-shell-extension-clipboard-indicator/tree/Gnome41

Captura de tela de 2022-01-14 11-42-51

rasyidrafi:
Why did this doesnt merged to master branch

Tudmotu:
@rasyidrafi unfortunately this doesn't work.

Tudmotu:
Closing. In the upcoming GNOME 45 version there will be image support (finally lmao).
https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/tree/image-support

PR #294, opened 2021-04-30
Do not create notification when history is cleared

vanitasvitae:
Since the user is actively choosing to clear the history, the subsequent notification is redundant.
On my system I have to actively cancel the notification, which is quite annoying.


Tudmotu:
You can turn off the notification in the settings:
image

PR #298, opened 2021-05-17
Add keyboard shortcut to toggle private mode

### LINKED ISSUE #176: [Feature Request] Keyboard Shortcut Key to Toggle Private Mode
Hi,

Firstly, thank you for a fantastic extension. I use it every day and it has been a life saver.

One thing which would help greatly is the ability to toggle Private Mode using a keyboard shortcut.

The use case would be for example as you go to log into a web-site and copy a password from your password manager it would be great to quickly toggle Private Mode on, enter password, and then toggle Private Mode back off.

Currently I do this with the menu item, but a keyboard shortcut would save time for an action that I preform many times a day.

Thanks again for the great work.

### LINKED ISSUE #186: [Feature Request] Set Private Mode default value
Please give the option to set default settings or persistent settings across restarts.

davekch:
This pull request addresses #176. I added two things:

 - a keyboard shortcut to toggle private mode
 - add private-mode to the schema so that it's editable from a script

This is useful when you're copying passwords from a password manager but don't want them to show up in the clipboard-manager. You can now either

 - use the shortcut to enable private mode before copying
 - if you copy from the terminal, you can now enable private mode by typing `gsettings --schemadir ~/.local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com/schemas/ set org.gnome.shell.extensions.clipboard-indicator private-mode  true` (only really useful in a script or as an alias tbh)

Since I don't have gnome shell 40 for testing, I made the changes on top of v34.

fixes #176, fixes #186


extension.js
```diff
@@ -846,6 +855,12 @@ const ClipboardIndicator = Lang.Class({
 
     _toggleMenu: function(){
         this.menu.toggle();
+    },
+
+    _togglePrivatemode: function() {
+        this.privateModeMenuItem.toggle();
+        this._settings.set_boolean(Prefs.Fields.PRIVATEMODE, this.privateModeMenuItem.state);
+        this._onPrivateModeSwitch();
```
SUPERCILEX, line 863: I don't think this is necessary... shouldn't `toggle` trigger `_onPrivateModeSwitch`?

SUPERCILEX (commented):
Please add `fixes #176, fixes #186` somewhere in your PR description.

SUPERCILEX:
Just tested it out and can confirm that everything works

extension.js
davekch: no, as far as I understand, `toggle` only moves the UI element. Without calling `_onPrivateModeSwitch`, the UI will look like it's in private mode, but it will still remember clipboard snippets

SUPERCILEX: Are you sure? I just tried commenting out the two lower lines and things seemed to still work as expected.

davekch: You are correct, those lines are redundant, not sure what I did wrong last time I checked. I removed them.

SUPERCILEX: Awesome, thanks! :)

davekch:
can I do anything more to get this merged?

SUPERCILEX:
@davekch I don't have merge rights unfortunately, but I'm working with the maintainer to figure something out.

SUPERCILEX:
@davekch See https://github.com/SUPERCILEX/gnome-shell-extension-clipboard-indicator/issues/1. I'm going to end up releasing my own version of the extension (which already has your PR merged). You can follow that issue to get an update on when it's in the store or install it from source.

schlagmichdoch:
@Tudmotu Why was this closed? I am interested in adding this functionality.

PR #304, opened 2021-08-27
do not save the history by default, and clarify the setting's label

attila-lendvai:
it's a major security issue to save people's clipboard to the disk
by default, and especially so if the name of this setting is not clear.

ideally this setting should be a dropdown box called _Persist history_:
 1) disabled
 2) favorites
 3) everything


Tudmotu:
1. Your PR has nothing to do with what you are suggesting
2. This is the opposite of the purpose of this extension
3. This extension is not supposed to be secure. If you are looking for a secure solution, you should find a different product

attila-lendvai:
> This extension is not supposed to be secure.

the concept of a clipboard does not entail in any way that its contents are freely accessible to others, especially not in a persisted way on the disk.

it's not like your extension is not secure... it's straight out covertly overriding a basic tenet of the clipboard. someone installs a clipboard *indicator*, and it ends up saving his passwords in cleartext on the disk. and even if they check, this fact is obfuscated.

so yeah, i guess you're right: backdoors are not supposed to be secure...

PR #313, opened 2021-11-05
build: support for gnome 41

Biplab-Dutta:
On updating to Fedora 35 which has gnome 41, the extension didn't work. This contribution allows the extension to work on gnome 41 too.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,6 +1,7 @@
 {
     "shell-version": [
-        "40"
+        "40",
+        "41"
     ],
     "uuid": "clipboard-indicator@tudmotu.com",
     "name": "Clipboard Indicator",


telometto:
Very nice if this gets implemented. This tool is a must-have for GNOME.

Biplab-Dutta:
> Very nice if this gets implemented. This tool is a must-have for GNOME.

@telometto Until an update is pushed, you can customize the extension to make it work on your machine. To do so, navigate to `~/.local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com/metadata.json`. Then change 
```
  "shell-version": [
     "40"
  ],
  ``` 
  to 
  ```
  "shell-version": [
    "40",
    "41"
  ],
  ```

PR #319, opened 2021-11-15
Update french locale

hjwk:
- Fix a stutter I made previously
- Fix a copy paste

### DIFF
--- a/locale/fr_FR/LC_MESSAGES/clipboard-indicator.po
+++ b/locale/fr_FR/LC_MESSAGES/clipboard-indicator.po
@@ -44,11 +44,11 @@ msgstr "Tout effacer ?"
 
 #: extension.js:340
 msgid "Are you sure you want to delete all clipboard items?"
-msgstr "Etes-vous sûr de vouloir effacer tout effacer ?"
+msgstr "Etes-vous sûr de vouloir tout effacer ?"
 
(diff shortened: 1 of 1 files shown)


locale/fr_FR/LC_MESSAGES/clipboard-indicator.po
```diff
@@ -44,11 +44,11 @@ msgstr "Tout effacer ?"
 
 #: extension.js:340
 msgid "Are you sure you want to delete all clipboard items?"
-msgstr "Etes-vous sûr de vouloir effacer tout effacer ?"
+msgstr "Etes-vous sûr de vouloir tout effacer ?"
 
 #: extension.js:341
 msgid "This operation cannot be undone."
-msgstr "Cette action ne peut ëtre annulée."
+msgstr "Cette action ne peut être annulée."
```
SUPERCILEX, line 51: Should we also add a `pas` here? So `ne peut pas ...`.

hjwk: Hmm, I wanted to keep it short so I omitted `pas` . But both forms are correct.

SUPERCILEX: Sounds good. :+1:

PR #331, opened 2022-01-09
Fix bug causing copy to be ignored

### LINKED ISSUE #290: Bug: Highlights wrong "active" clipboard item when favorites are involved
Hi, I'm using GNOME 40 (if that's relevant).

Steps to reproduce this bug:

1. Copy this to your clipboard: `one two three`
2. Open the menu and mark the `one two three` entry with a Star (favorite).
3. Copy this to your clipboard: `two`
4. Look in the menu. "two" will be set as active (dot) properly.
5. Click on the "one two three" favorited item in your menu to make it active instead.
6. Look in the menu. "one two three" will be set as active (dot) properly.
7. Copy this to your clipboard: `two` (yes it's the same word as before)
8. Look in the menu. "one two three" will be WRONGLY set as the active clipboard contents, even though contents is now "two" again.

Step 2 isn't necessary (marking as favorite) but it makes the bug more likely to happen. The bug can happen even if you don't do that.

Any ideas how to solve this? Hopefully it's a simple bug.

SUPERCILEX:
Fixes #290

Good ol' off-by-one error. It should either be `itemIndex <= registry.length - 1` or `itemIndex < registry.length`. I chose the latter.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,17 @@
+name: CI/CD
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '17'
+          cache: 'npm'
+      - name: Check style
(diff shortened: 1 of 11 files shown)


SUPERCILEX:

PR #332, opened 2022-01-11
[WIP: needs option to disable] Support paste-on-click

### LINKED ISSUE #109: FR: Directly paste from list

### LINKED ISSUE #153: suggestion
if I click on any text in the clipboard that means I want to paste it. if possible, can this extension directly put the text from the clipboard to the area where I am currently typing the text without pressing ctrl-v

### LINKED ISSUE #327: [Feature Request] Paste on selection
Currently it's already possible to navigate history and change clipboard selection with just a keyboard, however after older item is selected, one must still paste it again manually. It would be great to have Windows-like behaviour where one could paste on selection as well.

## Current Behaviour
1. Setup `Super+V` shortcut to bring up clipboard history
2. Invoke `Super+V`
3. Navigate with arrow keys to select older item
4. `CTRL+V` to paste that older item
5. Item gets pasted

## Desired Behaviour
1. Setup `Super+V` shortcut to bring up clipboard history
2. [New] Toggle setting to paste on item selection
2. Invoke `Super+V`
3. Navigate with arrow keys to select older item
4. Item gets pasted

SUPERCILEX:
Fixes #109, fixes #327, fixes #153

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,17 @@
+name: CI/CD
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '17'
+          cache: 'npm'
+      - name: Check style
+        run: npx prettier --check .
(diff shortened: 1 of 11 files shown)


SUPERCILEX:

PR #335, opened 2022-01-15
Tidy code for supporting images

SUPERCILEX:
@Tudmotu Ok, I addressed my own review feedback here. I've played around with it, and everything works now except for the aspect ratio issue. For that, I think we'd need to use something like this:
```js
const image = new St.Widget({
  style_class: 'ci-image'
});
image.add_child(St.TextureCache.get_default().load_file_async(
  Gio.file_new_for_path(buffer), -1, -1, 1, 1));
```

Wasn't able to get the sizing to work properly though. Do you want this fixed before merging your PR? I'm fine with opening a new issue and fixing it later.


SUPERCILEX:

Issue #342, opened 2022-02-20
Broken on Gnome 42 main / 42.beta


VitalyAnkh:
After building the latest gnome-shell and gjs, clipboard-indicator reports `Object 0x3acca7228850 is not a subclass of GObject_Object, it's a Object`. Maybe it's caused by the mozjs, which updates from version 78 to version 91.
![image](https://user-images.githubusercontent.com/24508452/154838664-b3f3ad17-66ab-4c8d-85ec-1c452f7693fd.png)

taivlam:
I'm also experiencing this issue on Debian Unstable with GNOME.

spiffytech:
openSUSE Tumbleweed has rolled out Gnome 42, so this is hitting me.

SUPERCILEX:
FYI, I've rewritten this extension and have been told it works on Gnome 42 (though I haven't been able to test it myself): https://github.com/SUPERCILEX/gnome-clipboard-history

taivlam:
Alright, I'll give the extension from @SUPERCILEX a try on Debian Unstable with GNOME.  (This extension still works on Pop!_OS, but it's probably because COSMIC is based on GNOME 3.38 and not GNOME 40+ - yet.)

BTW, Is there a way to get the contents of ` gnome-shell-extension-clipboard-indicator` before it stopped working?  It has some contents I'd like to save before migrating over to @SUPERCILEX's extension.

SUPERCILEX:
You'll need to copy `.cache/clipboard-indicator@tudmotu.com/registry.txt` to your new OS. Then just install the new extension and your data will be automatically migrated.

taivlam:
Thanks for the info!  Although I really should've said I'm first switching GNOME extensions before moving between OS's (from Debian Unstable/Sid to Pop!\_OS) - that'll come later.

If you're wondering whether you have to manually transfer your whether or not your clipboard data will have to manually transferred to SUPERCILEX's extension, you don't.

Upon a new installation, `gnome-clipboard-history` will automatically detect any prior clipboard history from `gnome-shell-extension-clipboard-indicator`.  Amazingly awesome!

SUPERCILEX:
Oh gotya, `~/.cache/clipboard-history@alexsaveau.dev/database.log` will be the one to copy then.

kdpuvvadi:
just installed clipnoard history from gnome extenstion store, old hostry from this just appead

neitsab:
[...]

PR #348, opened 2022-04-07
GNOME 42 support

ccat3z:
Register ClipboardIndicator using GObject.registerClass(). Fix https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/342. It seems that ClipboardIndicator cannot be defined with Lang.Class since gnome shell 42. But I'm not sure why.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -1,6 +1,7 @@
 const Clutter    = imports.gi.Clutter;
 const Config     = imports.misc.config;
 const Gio        = imports.gi.Gio;
+const GObject    = imports.gi.GObject;
 const Lang       = imports.lang;
 const Mainloop   = imports.mainloop;
 const Meta       = imports.gi.Meta;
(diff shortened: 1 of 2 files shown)


TiagodePAlves:
I'm using this version now and it's working without problems. I've installed it with:

```bash
$ rm -rf $XDG_DATA_HOME/gnome-shell/extensions/clipboard-indicator@tudmotu.com
$ git clone -b gnome-42 git@github.com:ccat3z/gnome-shell-extension-clipboard-indicator.git $XDG_DATA_HOME/gnome-shell/extensions/clipboard-indicator@tudmotu.com
```

Testing with:

```bash
$ gnome-shell --version
GNOME Shell 42.0
```

mlujann:
why was the branch deleted ?

extension.js
```diff
@@ -62,8 +62,7 @@ class ClipboardIndicator extends PanelMenu.Button {
         this._clearLabelTimeout();
         this._clearDelayedSelectionTimeout();
 
-        // Call parent
-        this.parent();
+        super.destroy();
```
rmader, line 65: Nit: typo in the commit message.

metadata.json
```diff
@@ -1,7 +1,8 @@
 {
     "shell-version": [
         "40",
```
rmader, line 3: Did you ensure everything works with 40/41 as well?

rmader (approved):
Looks and works great for me on 42!

metadata.json
Tudmotu: This is a good point. I will remove these, just in case. Changes like these are often not backward compatible with Gnome.

Tudmotu:
Amazing!!
Thank you! :pray: 
Will merge it soon along with a couple others.

Apologies for taking so long, only now updated to 42 on my system.

PR #350, opened 2022-04-16
Update metadata.json

Cicada313:
added 42 for gnome 42 compatibility


rmader:
See https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/348 instead

Issue #358, opened 2022-06-16
"Private mode" can be accidentally enabled while clearing items from history


openjck:
### Steps to reproduce

1. Copy a handful of things
2. Put your mouse over the trash icon for the topmost item in history
3. Rapidly click the mouse

### Actual result

At least on my screen, with my font size, resolution, etc. *Private mode* becomes enabled.

### Expected result

*Private mode* does not become enabled.

### Other notes

It seems that what is happening is that the *Private mode* toggle only listens for a "mouse up" event. When the last trash icon is clicked, the mouse ends up resting above the *Private mode* toggle. When the mouse button is released, a "mouse up" event is fired on the *Private mode* toggle and *Private mode* is therefore enabled.

This can be reproduced using other methods. Instead of clicking on trash can icons at all, open the clipboard indicator menu, press down on the mouse button anywhere in the menu, move the mouse to the *Private mode* toggle, and release the mouse button. *Private mode* will be enabled.

SUPERCILEX:
This how we fixed it: https://github.com/SUPERCILEX/gnome-clipboard-history/issues/66#issuecomment-1133495237

PR #365, opened 2022-09-28
fix: fixes issue where mouse events trigger Private Mode accidentally

rlueder:
This small PR fixes the issue described on #358


Tudmotu:
This doesn't seem to fix the issue when I test it :frowning_face: I can still toggle the "Private Mode" by press-and-holding anywhere, dragging the mouse on top of "Private Mode" toggle, and releasing.
Unfortunately I think the solution proposed in the original thread is incorrect.

rlueder:
Thanks for checking, I wasn't fixing that issue but the one where deleting the last item on the list accidentally activates Private Mode, see before and after captures below. From the original thread it seems the issue with holding the mouse button and moving to another element is an issue upstream with Gnome itself.

BEFORE FIX

clipboard_current

AFTER FIX

clipboard_fix

Tudmotu:
Oh wow, thank you for this explanation!
Apologies for misunderstanding :)

Tested and it indeed works. Thanks! :pray:

PR #366, opened 2022-09-28
fix: fixes wrong highlight of active clipboard item

rlueder:
Small PR to fix issue described on #290


Tudmotu:
Thank you :pray:

PR #374, opened 2022-12-20
Update metadata.json to run on Fedora 37

mchoma:
This is my local change to make it work on Fedora 37. So I thought I will share. I do not see if that is all what is needed.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,7 +1,8 @@
 {
     "shell-version": [
         "42",
-        "43"
+        "43",
+        "43.2"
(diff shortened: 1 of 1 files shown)


Tudmotu:
Hi,
Thank you for taking the time opening this PR.

I know I am very late to respond here, but I have one question -
Why is this necessary?
Personally I run GNOME 43.3 and have no issues.
Does Fedora have some special treatment of GNOME versions?
Do you have any resources explaining why this change is necessary?

PR #394, opened 2023-09-12
feat: pressing delete removes higlighted entry (#1) 

### LINKED ISSUE #375: Keyboard UI for deleting entries
Thanks for this useful extension.
Currently there's a keyboard UI to open the clipboard manager and to select the item to paste, however there is no keyboard option to remove items from the list.

It would be great if there was, so that keyboard heavy users could easily remove items.

I suggest that pressing the <kbd>Delete</kbd> key should remove the currently highlighted item, just as if the user had mouse clicked on the bin icon.

orpic:
1. Pressing the delete button when on a highlighted entry removes the entry
2. After the entry is removed the focus shifts to the next item and if empty then to the private

Tested it on GNOME OS Nightly ( via GNOME Boxes )

Fixes #375

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -386,6 +386,19 @@ const ClipboardIndicator = GObject.registerClass({
         }
     }
 
+    _findNextMenuItem (currentMenutItem) {
+        let currentIndex = this.clipItemsRadioGroup.indexOf(currentMenutItem);
+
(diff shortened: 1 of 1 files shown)


Tudmotu:
Thank you! :pray:

PR #396, opened 2023-09-15
bugfix: select next item after delete key event, favourite key event

### LINKED ISSUE #375: Keyboard UI for deleting entries

orpic:
Changes 
1. Finding the next available menu item  ( include extra cases )
2. The focus of the next item event comes before the actual delete or favorite key event

Fixes #375

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -400,11 +400,23 @@ const ClipboardIndicator = GObject.registerClass({
     _findNextMenuItem (currentMenutItem) {
         let currentIndex = this.clipItemsRadioGroup.indexOf(currentMenutItem);
 
-        for (let i = currentIndex +1; i < this.clipItemsRadioGroup.length; i++){
-            let menuItem  = this.clipItemsRadioGroup[i];
-            if(menuItem.actor.visible) {
+        // for only one item
+        if(this.clipItemsRadioGroup.length === 1) {
+            return null;
+        }
+
+        // when focus is in middle of the displayed list
+        for (let i = currentIndex - 1; i >= 0; i--) {
+            let menuItem = this.clipItemsRadioGroup[i];
+            if (menuItem.actor.visible) {
                 return menuItem;
             }
+        }
+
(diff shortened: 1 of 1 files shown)


orpic:
I have placed the focus action first before the delete or the favorite key press event because after these events the array has some changes and is difficult to identify the next item after the item itself is not present.

Let me know if anything seems out of place.

schlagmichdoch:
@orpic does this also work when deleting the first item of a longer list? I don’t find that special case in your code.

orpic:
@schlagmichdoch  first item from the bottom?

schlagmichdoch:
Ah so it's the other way round then. Alright! Just wanted to make sure you thought of all cases (top item, bottom item, middle item, only item in list). Thanks for implementing!

Tudmotu:
Nice! Thank you :slightly_smiling_face: 

I think it still has some issue with the image entries:



IMO, this is still better than current behavior so I am merging it cause I want it in the version before GNOME 45.
Feel free to open another PR if you'd like to fix the behavior with images, but worst case I will fix it some when in the future.

orpic:
Are image entries in some other array than  `clipItemsRadioGroup`?

Tudmotu:
@orpic They should be, yes. And you can see in the gif it does work sometimes.
Looking at your code I think maybe the behavior I'm seeing is not related to the images specifically.
But I'm not sure what it is. I believe your code should work :thinking: 

I think for now we can leave it like that, and I'm sure once GNOME 45 is out I will be getting bug reports with different scenarios. It will help us figure out what is going on there.

PR #398, opened 2023-09-20
Disable notifications for "Do Not Disturb" and "Private" modes

xalt7x:
Simple port of [commit from the Clipboard History](https://github.com/SUPERCILEX/gnome-clipboard-history/commit/7e22dc026c73ed4a27b447529c4fae5cb6ec74c2) extension.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -788,6 +788,14 @@ const ClipboardIndicator = GObject.registerClass({
     }
 
     _showNotification (message, transformFn) {
+        const dndOn = () =>
+          !Main.panel.statusArea.dateMenu._indicator._settings.get_boolean(
+            'show-banners',
+          );
+        if (PRIVATEMODE || dndOn()) {
+          return;
+        }
+
         let notification = null;
 
         this._initNotifSource();


Tudmotu:
Thanks @xalt7x. I can't reproduce this. I don't see notifications in these situations.

xalt7x:
Hello @Tudmotu 
Yes, I've tested it.
Please take a look at my screencast and explanation below.
I'm using "Previous" / "Next entry" keyboard shortcuts to switch entries.
Here's what happens in both cases:
1. "Private mode"
Entries are hidden in the indicator but still visible in notifications when I switch with shortcuts
2. "Do not Disturb"
Notifications are naturally hidden by GNOME, but they still appear in a Message Tray.
And unlike in normal mode (when they disappear pretty quickly), in DnD mode they get stuck there until the user clears them. It's also inconvenient with extensions that add more precise information about unread messages (e.g. "[Notification Counter](https://github.com/vkrizan/NotificationCounter)")
[Screencast from 2023-09-20 14-41-29.webm](https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/assets/44649402/c473fa84-670d-46aa-b6ec-420d5c0b5670)

Tudmotu:
Thanks @xalt7x. I will look into it. It seems you are using an old version of the extension, are you sure you are on `gnome45` branch?

Also, your report made me find a different bug so thanks for that as well :smile:

Issue #293, opened 2021-04-29
Option to hide clipboard icon


Freeplayg:
Since you can set a keybind to toggle the popup, maybe an option to hide the icon would to be nice so that it'll only be there when it's needed.

3m4r5:
same here, I don't need the icon

Tudmotu:
Interesting idea :+1:

ryleu:
Might work on this.

Tudmotu:
Need to also think how users will access the settings if the icon is hidden.

3m4r5:
@Tudmotu from the gnome extensions app?

3m4r5:
or using the the keyboard shortcut to invoke the panel & the clicking on the settings option

Tudmotu:
@3m4r5 ah yes, forgot about that

PR #423, opened 2023-11-28
Implemented option to hide clipboard icon completely. See issue #293

markocic:


Tudmotu:
Thank you!
It might take a bit before I release this change on e.g.o because I am currently still trying to fix all the mimetype issues.

Tudmotu:
Released in v57

PR #424, opened 2023-11-28
Implemented option to clear clipboard history on every boot. Fixes #372.

### LINKED ISSUE #372: Add an option to clear on every boot or similar
This would be useful for someone who brings the laptop around or shares the computer with other people or who just wants to have peace of mind. This could be disabled by default so it won't interfere with the normal operation of the extension.

markocic:
Closes: https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/372


markocic:
Glad to help! Saw your message in the IRC and thought I could help out a bit. Any updates on the mimetype issue?

Tudmotu:
@markocic Can you please make it so images are also deleted from the cache directory?
Images as saved as separate files using their hash value as filename. We need to delete these from the cache dir as well.

markocic:
@Tudmotu sure! I will push a commit soon to fix that.

Tudmotu:
Thank you :pray: 
Will test it out later today.

Tudmotu:
Released in v57

PR #429, opened 2023-12-26
Update clipboard-indicator.po

io-qar:
Update Russian locale


Tudmotu:

PR #435, opened 2024-01-21
Complete and correct Spanish translations

arturo-source:
I saw you added more texts to translate, so I added the translations.

Thank you for such a great Gnome extension.


Tudmotu:

PR #436, opened 2024-01-27
Update Chinese Simplified translation

hiiiha:
Translation update


Tudmotu:

PR #440, opened 2024-02-21
Update Italian language

ctrlVnt:
Just add some Italian strings


Tudmotu:

PR #450, opened 2024-04-03
Fix and update Italian locale

xalt7x:
Removes redundant backward slash in it.po as "make" reports error:

```
make compile-locales
...
locale/it/LC_MESSAGES/clipboard-indicator.po:140:34: invalid control sequence
msgfmt: found 1 fatal error
```

Also updates mo file.


Tudmotu:
Thank you (2) :pray:

PR #454, opened 2024-04-06
registry: don't trust the clipboard content type

### LINKED ISSUE #447: Doesn't work due to clipboard entry parsing
Appears the extension is trying to do something with the clipboard entry. I moved dir `/usr/share/devhelp/books/` and created a empty file `/usr/share/devhelp/books`, the extension loaded fine.

After clearing all clipboard entries, it loaded fine and all was ok.

I think clipboard clearing should work irrespective of other features.

### journalctl log:

```
Mar 30 18:12:30 linux gnome-shell[24359]: JS ERROR: Gio.IOErrorEnum: Error opening file /usr/share/devhelp/books: Is a directory
                                          fromJSON/bytes</<@file:///home/sid/.local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com/registry.js:232:47
                                          @resource:///org/gnome/shell/ui/init.js:21:20
```

### LINKED ISSUE #453: RCE vulnerability when copying full system paths
The entry parsing is broken, and I don't even know whether the underlying issue is the extension or something the extension depends on. It leads to a RCE, basically. If you copy a full system path, e.g `/usr/bin/ls` and the mimetype gets set to `UTF8_STRING` it will get **executed/opened on the system!** And the `contents:` key is replaced by that. For the example mentioned before about `/usr/bin/ls`, it's what you get:

https://gist.githubusercontent.com/Edu4rdSHL/20d05e03dfe4777eca76dba3c68d0f63/raw/587b6f38a62329e0d09d9bd992616195835a5b62/gistfile1.txt

It does look bad, and I'm not familiar with developing extensions to see where and why is this happening.

Regards,
Ed

andyholmes:
The clipboard content type may be a filename represented by a variety of mime-types, so double-check it is supported type before loading the contents.

closes #447
closes #453

### DIFF
--- a/registry.js
+++ b/registry.js
@@ -228,18 +228,32 @@ export class ClipboardEntry {
 
             let file = Gio.file_new_for_path(filename);
 
-            bytes = await new Promise((resolve, reject) => file.load_contents_async(null, (obj, res) => {
-                let [success, contents] = obj.load_contents_finish(res);
-
-                if (success) {
-                    resolve(contents);
-                }
-                else {
-                    reject(
-                        new Error('Clipboard Indicator: could not read image file from cache')
-                    );
+            const contentType = await new Promise((resolve, reject) => file.query_info_async(null, (obj, res) => {
+                try {
+                    const fileInfo = obj.query_info_finish(res);
+                    resolve(fileInfo.get_content_type());
+                } catch (e) {
+                    reject(e);
                 }
             }));
+
+            if (!contentType.startsWith('image/') && !contentType.startsWith('text/')) {
+                bytes = new TextEncoder().encode(jsonEntry.contents);
+            }
+            else {
+                bytes = await new Promise((resolve, reject) => file.load_contents_async(null, (obj, res) => {
+                    let [success, contents] = obj.load_contents_finish(res);
+
+                    if (success) {
+                        resolve(contents);
+                    }
+                    else {
+                        reject(
+                            new Error('Clipboard Indicator: could not read image file from cache')
+                        );
+                    }
+                }));
+            }
         }
 
         return new ClipboardEntry(mimetype, bytes, favorite);


andyholmes:
I think this is a reasonable fix, but you're the clipboard expert now :upside_down_face:

registry.js
```diff
@@ -228,18 +228,32 @@ export class ClipboardEntry {
...
-                else {
-                    reject(
-                        new Error('Clipboard Indicator: could not read image file from cache')
-                    );
+            const contentType = await new Promise((resolve, reject) => file.query_info_async(null, (obj, res) => {
+                try {
+                    const fileInfo = obj.query_info_finish(res);
+                    resolve(fileInfo.get_content_type());
+                } catch (e) {
+                    reject(e);
```
Tudmotu, line 236: What will be the value of `contentType` if the promise is rejected?

andyholmes: Oh, it will effectively propagate the error from `query_info_async()` and `fromJSON()` will then throw an error.

Tudmotu: Ah yes, got it. Thanks :slightly_smiling_face:

ken-kuro:
Sorry to bring this up again. But since #460 was merged, I think we should discuss more about this, cause in the current code, if the `contentType` is [unknown](https://gjs-docs.gnome.org/gio20~2.0/gio.fileinfo#method-get_content_type), we're loading its content which might lead to RCE in some case. Do you think some whitelisting like this would be better ? And could you please explain why are we load the content with `contentType` `/text`? @andyholmes 

```javascript
const allowed = ['/image', '/text']
if (contentType && allowed.some(type => contentType.startsWith(type))  {
    bytes = await new Promise((resolve, reject) => file.load_contents_async(null, (obj, res) => {
        let [success, contents] = obj.load_contents_finish(res);

        if (success) {
            resolve(contents);
        }
        else {
            reject(
                new Error('Clipboard Indicator: could not read image file from cache')
            );
        }
    }));
}
else {
    bytes = new TextEncoder().encode(jsonEntry.contents);
}
```

andyholmes:
I can not explain that, sorry. The intention was to retain the same behavior, without relying on the advertised clipboard mime-type being accurate.

If the content type of the file is unknown, it is loading the literal contents of the clipboard; which is a file path.

andyholmes:
Sorry, I didn't review your previous pull request, I only read your explanation. You're right that this line:

https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/blob/72f779d974a1990eddafcd211f354e038a39e6be/registry.js#L240

Should instead be something like:

```js
 if (!contentType?.startsWith('image/') && !contentType?.startsWith('text/')) {
```

There's a couple ways you can perform the same logical test, but the branch should be only be followed if the content type is `text/*` or `image/*`.

Tudmotu:
1. I agree using the optional chaining as Andy suggests is probably the way we should have done this. I pointed out the issue in the PR but didn't have time to make any changes yesterday and wanted to get the fix out
1. More broadly speaking I think the original issue is the fact I used the "content" field to save the file path. In hindsight that was the wrong decision. If the file path would have been saved in a different field (e.g. `filePath`) the original problem would not exist

ken-kuro:
I see, at the end of the day, it's just how we want to deal with `null` or `undefined` content type, cause tbh, I'm not sure what exactly the case that it has those values , the docs just use [unknown](https://gjs-docs.gnome.org/gio20~2.0/gio.fileinfo#method-get_content_type) without a clear explanation. But if the `content` field contains the file path, I think we're fine with both approaches

andyholmes:
Unknown in this context is pretty literal. There are only so many ways you can safely detect the content of file (i.e. [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))).

PR #463, opened 2024-04-23
Updated German translation

normanhenges:
The German translation was updated to support the new settings like "Behavior" and fix some minor translation errors.


Tudmotu:
Thank you :pray:

Issue #459, opened 2024-04-09
"Text will be here" is displayed in the tray and nothing happens by clicking on it.


BuZZ-dEE:
"Hier erscheint der Text" or "Text will be here" is displayed in the tray and nothing happens by clicking on it.

OS: Arch Linux x86_64
DE: Gnome 46 on Wayland
Extension version: 61

sounddrill31:
+1

> "Hier erscheint der Text" or "Text will be here" is displayed in the tray and nothing happens by clicking on it.
> 
> OS: Arch Linux x86_64 DE: Gnome 46 on Wayland Extension version: 61

OS: OpenSUSE Tumbleweed with 6.8.4
DE: Gnome 46 with wayland
Extension Version: 61

MeirKriheli:
@sounddrill31 Thanks, deleting the cache and re-enabling the extension helped here as well.

Tudmotu:
Possibly related to this?
https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/458

Tudmotu:
Can everyone here please try updating to version 62 and see if it fixes the issue for you?

Thanks! :pray:

tanjilbhuiyan:
Thanks, it seems the issue is fixed. But faced a weird issue. i uninstalled previous one. then installed again. Got error in Gnome extension page. though the extension was downloaded. Had to log out and log in again. It seems working now.

Tudmotu:
Thank you for testing it out @tanjilbhuiyan !

To update the extension you must log out and back in. This is true every time you update — it only updates after the next login.

sounddrill31:
Works so far for me!

Thanks for the fix

arturo-source:
Hello @Tudmotu,
I had the same problem when I updated to 61, then I updated to 62 and the problem kept. 
Unfortunately I already deleted cache, as @sounddrill31 recommended and it started working again, so I cannot test anymore.

Tudmotu:
Thank you for confirming @sounddrill31 !

@arturo-source apologies you had to remove your cache file 😕 Hope this won't have to happen again but if it does, next time I recommend renaming the file instead of deleting it, so that you can restore it later on.

MeirKriheli:
Same here: removed the cache (was not that important anyway), so can't check.

DerPate:
@Tudmotu can confirm the update to Version 62 fixed the problem for me

[...]

PR #480, opened 2024-06-19
Fix UTF8_STRING in registry.txt renders extension unresponsive on Gnome 46 (Wayland)

Snorch:
The idea of the fix comes from @gentoo-root, @AmeyaVS and @alephnull
comments in the original issue:

Changing "UTF8_STRING" to "text/plain;charset=utf-8" in registry.txt
seems to be fixing the problem after reloading the extension.

So let's use this.isText() instead of mimetype.startsWith('text/')) in
fromJSON text checking.

Without this patch "Text will be here" appears in the tray and extension
is unresponsive.

Fixes: #459

### DIFF
--- a/registry.js
+++ b/registry.js
@@ -214,12 +214,18 @@ export class ClipboardEntry {
         return Uint8Array.from(contents.match(/.{1,2}/g).map((byte) => parseInt(byte, 16)));
     }
 
+    static __isText (mimetype) {
+        return mimetype.startsWith('text/') ||
+            mimetype === 'STRING' ||
+            mimetype === 'UTF8_STRING';
+    }
+
     static async fromJSON (jsonEntry) {
         const mimetype = jsonEntry.mimetype || 'text/plain;charset=utf-8';
         const favorite = jsonEntry.favorite;
         let bytes;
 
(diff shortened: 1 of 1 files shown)


alephnull (approved):

Tudmotu:
Thank you for this, will test it out and release in the next few days 🙏

TriplEight:
I faced this bug with "Text will be here" and unresponsive extension v62. Then I found fixes https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/459#issuecomment-2044686278 and https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/459#issuecomment-2138858968.

So I first checked what was in `~/.cache/clipboard-indicator@tudmotu.com/registry.txt` and it was empty! The bug is reproducible in this state. The workaround would be: 
- remove the empty `~/.cache/clipboard-indicator@tudmotu.com/registry.txt`
- restart the extension

To improve the fix I'd recommend checking/restoring the structure of `registry.txt`.

Tudmotu:
Apologies for the delay. Pushed to e.g.o. Thank you! :pray:

Issue #495, opened 2024-09-30
Chance of making the history larger than 200


luisalvarado:
![image](https://github.com/user-attachments/assets/61962328-c5b8-4d4e-bec2-d4113ccd390b)

Could there be a chance of making the History Size much bigger than 200, for example up to 10K. I understand the memory consumption of it because of images and other factors, but I would really like to have it much larger and maybe the user could be advice about the memory usage if larger than that?

I use this tool a lot on a daily basis and thank you for it because it makes me more productive in several scenarios but that 200 is very limiting and something much above 1000 would be better for my day to day life. So was wondering if this could be done. Thank you.

Tudmotu:
Sure, I'd accept a PR raising the max to 10,000.

Tudmotu:
Uploaded to e.g.o, limit will be 10k

PR #497, opened 2024-11-05
Increase History and Max cache file size

abdurrahmanseyidoglu:


prefs.js
```diff
@@ -29,7 +29,7 @@ class Settings {
             title: _("History Size"),
             adjustment: new Gtk.Adjustment({
                 lower: 1,
-                upper: 200,
+                upper: 1000,
```
banocean, line 32: @Tudmotu in https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/495 said he will accept raising the limit to 10k not 1k

banocean (changes requested):
everything besides that looks ok

abdurrahmanseyidoglu:
I didn’t increase it to 10K because I was worried it might affect the performance of the GNOME desktop environment on low-end systems.

1K seemed more reasonable to me. 😊

PR #506, opened 2025-01-31
Fix system lag on copy (WIP)

### LINKED ISSUE #428: clipboard-indicator causes system lag while copying
After one of the last updates I noticed that clipboard-indicator causes a system lag when pressing `CTRL + C`.

Now every time I copy something the system hangs. How depends on the size of the "thing" (length of text, size of image, etc) added to the clipboard.

A JPG with one or two MB causes a hang for more than 1 second.

I had to turn the clipboard-indicator off, because of that.

Using manjaro (arch) + wayland

danielhuang:
Not yet fully tested, should fix #428

### DIFF
--- a/registry.js
+++ b/registry.js
@@ -29,14 +29,9 @@ export class Registry {
 
             registryContent.push(item);
 
-            if (entry.isText()) {
-                item.contents = entry.getStringValue();
-            }
-            else if (entry.isImage()) {
-                const filename = this.getEntryFilename(entry);
-                item.contents = filename;
-                this.writeEntryFile(entry);
-            }
+            const filename = this.getEntryFilename(entry);
+            item.contents = filename;
+            this.writeEntryFile(entry);
         }
 
         this.writeToFile(registryContent);
@@ -229,7 +224,7 @@ export class ClipboardEntry {
         const favorite = jsonEntry.favorite;
         let bytes;
 
-        if (ClipboardEntry.__isText(mimetype)) {
+        if (false) {
             bytes = new TextEncoder().encode(jsonEntry.contents);
         }
         else {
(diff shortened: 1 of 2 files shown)


registry.js
```diff
@@ -148,9 +149,19 @@ export class Registry {
...
+        if (this.hashmap.has(entry)) {
+            return this.hashmap.get(entry);
+        }
+        const hash = entry.asBytes().hash();
+        this.hashmap.set(entry, hash)
+        return hash
+    }
 
     getEntryFilename (entry) {
-        return `${this.REGISTRY_DIR}/${entry.asBytes().hash()}`;
```
Tudmotu, line 153: Omg was this the problem all along?

Tudmotu: I think we can also try using just some random string, since the filename is saved in the json anyway.

danielhuang: I'm not completely sure — I only get lag on copy after several days of uptime, so I end up having a multi-day edit/test cycle.

I'll report back after further testing.

danielhuang: After testing it some more, copying large images still causes freezes on copy (up to several seconds), and copying text causes short freezes (270ms).

danielhuang: It might be best to not compute the hash at all (or at least not on the main thread), since the first copy of a large image will always require computing the hash even if it's cached.

danielhuang:
Currently the code is removing checks for content, since as the history size grows, every new copy will have to check the entire history to check for duplicates.

danielhuang:
A real fix would probably require having some way to calculate hashes off of the main thread, instead of blocking.

Tudmotu:
@danielhuang apologize for the delay in response.
So this did not solve the issue unfortunately?

Issue #500, opened 2024-12-03
Previous Entry and Next Entry Should not use gnome notifications/alerts


somtooo:
Previous Entry and Next Entry Should not use gnome notifications/alerts because now that they are coupled any setting for gnome alerts affects when toggling thru next entry. 
e.g gnome alerts timeout after 3s and if i have a long list of entries to go thru the window will close before i even get to my desired entry.

tesar-tech:
I am with you on this one.
At least there should be settings to turn the notification off.

somtooo:
@tesar-tech the notification style is cool makes it easy to cycle thru entries but it should be its own window. @Tudmotu  is this how it was meant to operate?

tesar-tech:
> [@tesar-tech](https://github.com/tesar-tech) there is an option to turn notifications off in the settings.

Not for "entries movement", just for history and copy...

![Image](https://github.com/user-attachments/assets/30a0549f-32b2-45b6-807f-ddcd7243818b)

Tudmotu:
Ah, gotcha @tesar-tech .
I would accept a PR that adds an option like that, if anyone wants to give it a try.

capaj:
I think this can be closed as the PR got merged and it was released @Tudmotu

FurkanTokmak:
@capaj I think issue still holds, #519 adds an option to turn off the notification on cycle but still when I navigate through long list the notification disappears mid navigation.

PR #519, opened 2025-03-20
Add settings for turning off notifications when cycling through the clipboard entries using hotkeys

### LINKED ISSUE #469: shows sensitive information on next and previous entries
I often put my totp, password and username from my password manager in clipboard and then
use the next entry hotkey for pasting the relevant information during login.

However each time I press the next entry hotkey I get a notification with that data shown:

image

So when doing screensharing people watching the screen / recording the session can read this data...
So I am not using the next previous functionality because of this.
A simple copy paste does not show it. 

I believe it is only when the extension tries to change the clipboard contents.

I have notification off

image

I am on latest master version as the V60 doesn't even allow me to use prev next.

CaragC:
This fixes #469 and [this](https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/500#issuecomment-2625210996) comment made on another issue

> [!IMPORTANT]
> The toggle is on by default, so the behaviour of the extension doesnt change witout setting the new preference

#### Preview of the preferences UI
image


CaragC:
just noticed this also fixes #469

Issue #407, opened 2023-10-25
Ignore input from certain programs (e.g. password managers)


Anuskuss:
Sometimes I have to manually copy something out of my password manager and it's a bit of a security concern if that shows up in the clipboard manager. I know that there's a private mode but I usually forget to turn that on before copying.

One way to tackle this would be to check the active window (title) when a copy was issued. Or in the case of KeePassXC automatically enter private mode when the window exists (it hides in the tray when not in use).

Tudmotu:
Yes, this is a highly requested feature, but unfortunately I have very limited time to work on this extension and last time I checked it was unclear how to even implement this.

If you think you might know how to implement it, I would gladly accept a PR. Let me know if you wanna work on this.

CaragC:
What I am currently trying is something close to [Panos](https://github.com/oae/gnome-shell-pano) implementaion, where the extensions looks at the currently focused window, and if the exclusion list, defined by the user in the settings, contains the app, the entry is not added

Anuskuss:
@CaragC I think filtering the `x-kde-passwordManagerHint` MIME type (like is done in [gnome-clipboard-history](https://github.com/SUPERCILEX/gnome-clipboard-history)) is way easier. Your password manager gotta support that though.

CaragC:
I opened a [PR](https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/518) implementing it like pano does, I think it works pretty well, but I would be open to change it or maybe add another toggle on top for enabling the MIME type filtering

Anuskuss:
That certainly looks interesting, and I actually prefer it over MIME type because it's more versatile. But to be honest I no longer use `gnome-shell-extension-clipboard-indicator`. Maybe I'll check it out once the PR is merged (although `gnome-shell-pano` looks interesting as well).

sith-on-mars:
Thanks for the PR. How do I enable this function for flatpak apps? I put `flatpak run org.gnome.World.Secrets` and `org.gnome.World.Secrets`, which neither of them worked.

CaragC:
hmm wierd, when i use `org.gnome.World.Secrets` it stops copying from secrets. Is the window in focus while you are copying? If you want to you can try to press `alt+f2` then type `lg` and then go to the `windows` section, there you should see Secrets and under the title it should say `wmclass`. This should be what you need to type in to get it to work

PR #528, opened 2025-04-01
Add button when adding excluded apps for searching all applications

CaragC:
The most important changes is the addition of a popover for selecting installed applications and a search functionality within the popover. The selected applications id, which will also be the wmclass of the apps window, will then be filled into the inout field. This is in response to [this](https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/407#issuecomment-2761613436) comment

#### Preview
image
 ---
image

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -340,23 +340,142 @@ class Settings {
     }
 
     #createExcludedAppInputRow() {
-        const entry_row = new Adw.ActionRow();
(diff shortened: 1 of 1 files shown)


Tudmotu:
The check icon was broken on 49 and the input size was incorrect because of the ActionRow `title` element, so I set it to visible = false.
image

PR #529, opened 2025-04-02
Made options to hide search bar, private mode, clear, and/or settings buttons & fixed Paste on Select bug.

JaredTweed:
Did title and also made pressing Enter and clicking on an item have consistent behaviour. Fixed glitch where Paste on Select only worked when pressing Enter, and not when clicking.

Please implement this change, it is really important to me, I have been using this version of the clipboard manager for a couple weeks.

My changes are well commented and should be easy to understand.

This resolves issues [#515](https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/515) and  [#516](https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/issues/516).

Now using the settings it can look like this:
image
image

PR #534, opened 2025-05-08
Made options to hide search bar, private mode, clear, and/or settings buttons & fixed Paste on Select bug.

JaredTweed:
This pull request is built on top of my other pull request. 

This ensures that:
* Text always pastes.
* The text that pastes is always the selected item, and never accidentally pastes the previously selected item.

Both of those bugs were frequent especially on slower hardware, but I think I entirely fixed those errors.


extension.js
```diff
@@ -66,6 +70,9 @@ const ClipboardIndicator = GObject.registerClass({
     GTypeName: 'ClipboardIndicator'
 }, class ClipboardIndicator extends PanelMenu.Button {
     #refreshInProgress = false;
+    _pastePending = null;
+    _previewLaterId = 0;
+    preventIndicatorUpdate = false
```
Tudmotu: Why not private vars?

extension.js
```diff
@@ -700,9 +758,20 @@ const ClipboardIndicator = GObject.registerClass({
...
-            this._refreshIndicator();
+        if (selectionType !== Meta.SelectionType.SELECTION_CLIPBOARD)
+            return;
+        
+        // Step 3 — clipboard owner is now the one we set in #pasteItem()
+        if (this._pastePending) {
+            const { previous } = this._pastePending;
+            this._pastePending = null;
+                this._keyboardPaste();            // paste immediately
+                this.preventIndicatorUpdate = false;
```
Tudmotu: Why did you have to move unsetting the `preventIndicatorUpdate` here?

extension.js
```diff
@@ -1082,70 +1145,68 @@ const ClipboardIndicator = GObject.registerClass({
...
-                    that._showNotification(index + ' / ' + menuItems.length + ': ' + menuItems[i].entry.getStringValue());
-                }
-                if (MOVE_ITEM_FIRST) {
-                    that._selectEntryWithDelay(menuItems[i]);
-                }
-                else {
-                    that._selectMenuItem(menuItems[i]);
-                }
+                i = (i + 1) % menuItems.length;
+                this._showNotification(/* … */);
```
Tudmotu: 1. Why did you remove the option to disable the notification?
2. As discussed in the PR comments, delay should not be removed
3. Honestly, the fact that you removed the notification text, and put a comment for some reason, indicates that you did not put in the necessary effort for testing your changes for possible regressions

extension.js
```diff
@@ -1082,70 +1145,68 @@ const ClipboardIndicator = GObject.registerClass({
...
-        if (this._setFocusOnOpenTimeout) clearTimeout(this._setFocusOnOpenTimeout);
-        if (this._pastingKeypressTimeout) clearTimeout(this._pastingKeypressTimeout);
-        if (this._pastingResetTimeout) clearTimeout(this._pastingResetTimeout);
-    }
+        for (const id of [
+            this._setFocusOnOpenIdleId,
+        ]) {
+            if (!id)
+                continue;
+            const laters = global.compositor.get_laters();
```
Tudmotu: Would this not throw an error in GNOME<44?

extension.js
```diff
@@ -1082,70 +1145,68 @@ const ClipboardIndicator = GObject.registerClass({
...
+                Meta.Laters.remove(laters, id);  // GNOME 44 +
+            } else { GLib.source_remove(id); }     // Idle fallback    }
+        }
 
+        if (this._previewLaterId) {
+            const laters = global.compositor.get_laters?.();
+            laters && Meta.Laters.remove
+                ? Meta.Laters.remove(laters, this._previewLaterId)
+                : GLib.source_remove(this._previewLaterId);
+            this._previewLaterId = 0;
```
Tudmotu: Why 0 and not null like with other source ids?

extension.js
```diff
@@ -471,9 +531,7 @@ const ClipboardIndicator = GObject.registerClass({
                     break;
                 case Clutter.KEY_KP_Enter:
                 case Clutter.KEY_Return:
-                    if (PASTE_ON_SELECT) {
-                        this.#pasteItem(menuItem);
-                    }
+                    // Let the 'activate' event handle the paste (to prevent double pasting)
                     this._onMenuItemSelectedAndMenuClose(menuItem, true);
```
Tudmotu: How is this triggering the "activate" event? If `_onMenuItemSelectedAndMenuClose` triggers "activate", wouldn't it cause an infinite loop since the "activate" even handler also invokes `_onMenuItemSelectedAndMenuClose`?

JaredTweed:
Q1) Meta.Laters is Mutter’s frame-synchronized scheduler. It runs callbacks at compositor phases like AFTER_REDRAW, which is ideal for preview image swaps and avoids UI races that idle/timers can’t. The code falls back to GLib.idle_add() if Meta.Laters isn’t present, keeping compatibility across versions.

Q2) Yes, the original delay buffered against focus and relayout churn while the menu opened. In #529 the focus delay is preserved using GLib.idle_add() so it doesn’t fight Shell’s animation, but the separate selection delay is removed and replaced with synchronous selection. This prevents the paste-ordering races caused by timers, and #534 then fully replaces them with event-driven sequencing.

Q3) The pasting fix can not be split, because the way it loads the menu causes it to load differently, causing some issues in the timers and the order in which they load things. #529 introduced the unified activation path and the _pastePending/owner-change hooks. #534 builds on this by replacing the last timeouts with sequential, signal-driven logic and fixing the “previous vs current selection” problem. While it could be split, #534 is a direct continuation of #529 and simpler to test together.

Q4) I had it on my last computer (a very old slow computer), where it would occasionally paste the second most recent instead of the most recent code (only when I copied large sections of text. Paste-on-select also failed when clicking an entry, because the paste is attempted before the clipboard owner switches and before focus returns to the target app. With #529 and #534, activation is unified: the menu closes first to return focus, the clipboard is updated, and the paste only happens once the owner-change signal is seen. This guarantees correct ordering and consistent behavior for both click and Enter. On my newer computer I don't have that issue.

JaredTweed:
The other benefit of making it run sequentially, is that it will be more reliable, and not rely on specific animation/loading speeds.

JaredTweed:
If you want, I can modify PR #529 to evade the same bugs as #534, but avoid using meta.laters instead of glib (so as to keep the change minimal). Would you like me to do that?

Tudmotu:
The delay has nothing to do with animations or layouts. You can see the delay is only used when the "move item to top" setting is enabled. This is because otherwise, iterating through the list of items via shortcuts becomes impossible: every time you select an entry it moves to the top, you select the next entry it moves to the top, the previous entry is now in place 2, you move to the next item and it's actually the previous entry, so you are effectively stuck in a loop between 2 entries. This delay should not be removed as it is crucial.

This PR has some issues which makes me uncomfortable merging it. I am leaving a review, but I highly suggest that you try first submitting a smaller PR, that does not include any fundamental changes to the way this extension works. Thousands of people if not more use it every day, and if it was this broken then I'd hear about it.

The scenario you are describing with the paste-on-select window focus race condition might present an actual issue. If that's the case, I suggest you find a reproducible example. I think maybe a moving the paste action to the "activate" handler is all that is needed (not sure about it). In any case, for me to be able to test a fix like that, I need to be able to first reproduce the underlying issue.

I know you put a lot of effort into this PR and I don't want to discourage you. But I have to consider the many many many users of this extension when I merge PRs, especially ones that make fundamental changes to the code. Thank you.

Tudmotu (changes requested):

extension.js
JaredTweed: _onMenuItemSelectedAndMenuClose doesn’t emit "activate" at all. it just updates selection state and closes the menu. The "activate" signal is emitted by PopupMenuItem itself when the user clicks or presses Enter. Before, I also called _onMenuItemSelectedAndMenuClose manually on Enter, which duplicated the effect. I’ve removed that (in a commit i will push in an hour or so) and now rely solely on the built-in "activate" handler, so there’s no loop and both click and Enter go through the same path.

extension.js
JaredTweed: Okay, in an upcoming commit i have fixed that. I will push it in an hour or so. I have made it use GLib instead of Meta.Laters

extension.js
JaredTweed: Good point! I'll make them private in an upcoming commit.

extension.js
JaredTweed: I used 0 because GLib source IDs are numeric and 0 is the standard sentinel for ‘no source.’ For signal handler IDs I use null since those aren’t numeric. Functionally both work in GJS, but 0 keeps the type consistent with what GLib.idle_add returns.

extension.js
JaredTweed: Shoot, I'm sorry, that's embarrassing. That will be fixed in the next commit.

extension.js
JaredTweed: I moved unsetting #preventIndicatorUpdate into _onSelectionChange() because that’s the first point we know our clipboard write has actually taken effect. Clearing it earlier risks a race where the indicator refreshes in between our write and the owner-changed signal, causing flicker or stale content; doing it on owner-changed ensures we only re-enable updates once our own write is acknowledged.

JaredTweed:
Okay, I did all the changes I found to be necessary and did a thorough bug search as well. Is this code sufficient? You can ignore PR #529 now and just pull this instead.

JaredTweed:
Sorry for not checking it better before.

JaredTweed:
This is the smallest I know how to make this PR, I am sorry. Thank you so much for the work you have put into this project!! I really appreciate your time.

constants.js
dynobo, line 1: I'm just following this topic, but I guess leaving the reformatting of this mapping out of the PR would help keeping the changeset small and easier to review? What do you think?

JaredTweed: Good idea! I just made that change.

JaredTweed:
@Tudmotu, I made the changes as clean as possible for you. Is it clean enough to be added to the extension?

JaredTweed:
@Tudmotu I made the PR much smaller now, and kept your version of the timeouts that you like so that there is no longer that risk of making it create breaking changes. I also updated the code so that it has no conflicts with the base and can now be merged with the click of a button. I hope these additions are sufficient.

JaredTweed:
@Tudmotu I also got rid of all that Meta.Laters stuff and all those unnecessary changes that I made before

extension.js
```diff
@@ -1356,7 +1409,8 @@ const ClipboardIndicator = GObject.registerClass({
 
             this._pastingResetTimeout = setTimeout(() => {
                 this.preventIndicatorUpdate = false;
-                this.#updateClipboard(currentlySelected.entry);
+                if (currentlySelected && currentlySelected.entry)
+                    this.#updateClipboard(currentlySelected.entry);
```
Tudmotu, line 1433: In what situation is `currentlySelected` not defined?

prefs.js
```diff
@@ -185,6 +185,16 @@ class Settings {
...
+        // Toggleable UI rows
+        this.field_show_search_bar = new Adw.SwitchRow({title: _("Show Search Bar")});
+        this.ui.add(this.field_show_search_bar);
+        this.field_show_private_mode = new Adw.SwitchRow({title: _("Show Private Mode")});
+        this.ui.add(this.field_show_private_mode);
+        this.field_show_settings_button = new Adw.SwitchRow({title: _("Show Settings Button")});
+        this.ui.add(this.field_show_settings_button);
+        this.field_show_clear_history_button = new Adw.SwitchRow({title: _("Show Clear History Button")});
+        this.ui.add(this.field_show_clear_history_button);
+
```
Tudmotu: Why do you create the fields and add them here instead of following the pattern used throughout the rest of the file where the fields are created on top and are only attached here?

extension.js
```diff
@@ -1151,17 +1202,19 @@ const ClipboardIndicator = GObject.registerClass({
...
             this._updateTopbarLayout();
             that.#updateIndicatorContent(await this.#getClipboardContent());
 
             // Bind or unbind shortcuts
-            if (ENABLE_KEYBINDING)
-                that._bindShortcuts();
-            else
-                that._unbindShortcuts();
+            if (ENABLE_KEYBINDING) that._bindShortcuts();
+            else that._unbindShortcuts();
```
Tudmotu: Meaningless whitespace changes like these clutter your PR. Please undo this change unless it is meaningful somehow. Avoid any changes that do not affect functionality.

extension.js
```diff
@@ -1135,6 +1180,12 @@ const ClipboardIndicator = GObject.registerClass({
         NEXT_HISTORY_CLEAR          = settings.get_int(PrefsFields.NEXT_HISTORY_CLEAR);
         CASE_SENSITIVE_SEARCH       = settings.get_boolean(PrefsFields.CASE_SENSITIVE_SEARCH);
         REGEX_SEARCH                = settings.get_boolean(PrefsFields.REGEX_SEARCH);
+
+        // NEW toggles
```
Tudmotu: What does "NEW" mean here? If someone sees this code in 3 years, would these still be "NEW"?
Please remove this comment and just maintain the same code style and patterns of the surrounding code.

extension.js
```diff
@@ -504,35 +532,49 @@ const ClipboardIndicator = GObject.registerClass({
...
         });
+
         menuItem.actor.connect('key-press-event', (actor, event) => {
             switch (event.get_key_symbol()) {
                 case Clutter.KEY_Delete:
+                    if (!DELETE_ENABLED) return Clutter.EVENT_STOP;
                     this.#selectNextMenuItem(menuItem);
                     this._removeEntry(menuItem, 'delete');
-                    break;
+                    return Clutter.EVENT_STOP;
```
Tudmotu, line 571: Why do you need to return these `EVENT_STOP` consts?

extension.js
```diff
@@ -737,6 +781,11 @@ const ClipboardIndicator = GObject.registerClass({
             }
         }
 
+        // Ensure MOVE_ITEM_FIRST also applies when PASTE_ON_SELECT fast-path skips _refreshIndicator()
+        if (PASTE_ON_SELECT && MOVE_ITEM_FIRST && !menuItem.entry.isFavorite()) {
+            this._moveItemFirst(menuItem);
+        }
```
Tudmotu, line 798: Why do you need to manually move the item here?

extension.js
```diff
@@ -595,9 +637,11 @@ const ClipboardIndicator = GObject.registerClass({
...
         menuItem.actor.add_child(icoBtn);
         menuItem.icoBtn = icoBtn;
-        menuItem.deletePressId = icoBtn.connect('clicked',
-            () => this._removeEntry(menuItem, 'delete')
-        );
+        menuItem.icoBtn.visible = DELETE_ENABLED;
+        menuItem.deletePressId = icoBtn.connect('clicked', () => {
+            if (!DELETE_ENABLED) return;
+            this._removeEntry(menuItem, 'delete');
+        });
```
Tudmotu: Why did you re-enable support for this feature? It's long gone as far as I can see.

extension.js
```diff
@@ -200,45 +206,40 @@ const ClipboardIndicator = GObject.registerClass({
...
+                    if (currentItem) global.stage.set_key_focus(currentItem.actor);
+                } else if (SHOW_PRIVATE_MODE && that.privateModeMenuItem) {
+                    global.stage.set_key_focus(that.privateModeMenuItem.actor);
                 }
             }, 50);
         });
 
-        // Create menu sections for items
-        // Favorites
+        // Favorites section
```
Tudmotu: Undo every unnecessary change. I will not accept a PR that has any unnecessary changes. Keep you PR as clean as possible and adhere to the style and patterns used in the existing code.

extension.js
```diff
@@ -48,6 +48,12 @@ let NEXT_HISTORY_CLEAR        = -1;
 let CASE_SENSITIVE_SEARCH     = false;
 let REGEX_SEARCH              = false;
 
+// NEW: toggleable UI elements
```
Tudmotu: Remove the "NEW" comment here as well.

Tudmotu (changes requested):
I left another review. This PR is still too much of a mess. Follow the comments I left you and I will review it again afterwards. I'm not sure if you are using claude or if it's you that is adding all these unnecessary changes, but please stop. If you are using an agent to help, you need to double review its changes and make sure everything works properly.

One more thing: DO NOT resolve comments on your own. Only the reviewer should resolve comments. Resolving comments others leave makes it harder to see if you actually implemented the proper changes. This is basic open source etiquette.

dynobo:
@Tudmotu, as a grateful user of your extension who curiously follows this topic here (as I like the idea), I just want to take the opportunity to thank you for insisting on high quality standards and appropriate workflows. 👍
A clipboard manager deals with quite sensitive information, and the way you handle this here increases my trust in the extension even further! Thank you for your great work! ❤️

JaredTweed:
@Tudmotu I will be more careful with the changes. Sorry about resolving the comments. I will let you resolve those. Thank you so much for being willing to take the time to review my code, I really appreciate your discipline in that. I really hope my contributions make it into the final product. Let me know if you would like any more changes.

JaredTweed:
@Tudmotu Thank you for your patience with me. Just so you know, I am ready for you to review it!

extension.js
capaj: I think we may want to return Clutter.EVENT_STOP to prevent the event from propagating up the actor hierarchy — otherwise GNOME Shell may also handle it (e.g., close the popup or trigger other UI reactions).

Tudmotu:
Thank you @JaredTweed ! Apologies for the delay. I appreciate you working on this and improving the PR. Your feature will be released in the next version.

JaredTweed:
Thank you so much, @Tudmotu! I look forward to seeing it.

Tudmotu:
@JaredTweed there's a hold back in extension reviews, and since the next version clipboard indicator contains many features, it will probably take a while before it is released. If you like, you can use the `gnome-50` branch that includes your changes, so you can hide these UI elements for now.

PR #557, opened 2025-09-22
Updated .pot and fr .po/.mo

ClairDePrince:
Updated .pot and .po/.mo from latest release


bywkgexmkiltxh:
Hi,

After updating .pot file (template), it would be great to enter the following commands:

1. Change directory to extension's root directory
```bash
cd /path/to/extension
```

2. Extend .po files (translations) from re-generated .pot file for all languages. Doing this will not mess the translations and allow other translators to translate the new strings in their respective language
```bash
find locale -name "*.po" | while read -r po_file; do
  msgmerge --update "$po_file" clipboard-indicator.pot
done
```

Tudmotu:
@p-sage  Thank you for this contribution.

@bywkgexmkiltxh I believe we have a command `make update-po-files` that should do that. But thank you nonetheless for mentioning that!

PR #562, opened 2025-10-12
Optimize _clearHistory

SiriusCrain:
With clipboard history of 50+ (approximately) items attempt to clear it lead to visible (500-1500ms) lag, so I first gathering all items which needs to be removed, and then remove them in a faster way, also cache and UI updated once after clear is finished.

Tested locally, spawned few hundred random items to cliboard and clear was instant after changes


SiriusCrain:
@Tudmotu review pls

PR #570, opened 2025-11-18
Add option to blink top bar icon on copy

ttmx:
I've been using this extension for years now, and always thought notifications were too big and distracting, but I still wanted to know when I actually copied something.

I implemented the option for the icon in the top bar to blink when something is copied. 

Demo here, ignore other extensions:

https://github.com/user-attachments/assets/1ae8d130-7250-44a4-8f47-1bdcb0e53cb5

Settings here:
image

Thanks for the great extension, hope this is useful to others!


Tudmotu:
Hi @ttmx thank you for this PR. It is well structured and concise. Once thing that was missing was disconnecting the signals on `destroy()`. I plan on including this in the next release, but since e.g.o currently has a large backlog, and the next clipboard indicator version has many changes, it might still take a while before it is released.  For now, you can use the `gnome-50` branch that includes your change.

ttmx:
Thank you ! I'm not usually a gnome dev, would you want me to figure out how to add the .disconnect, or would it be okay if you handled it?
Don't mean to place undue burden on you.

Tudmotu:
@ttmx thank you. I already handled it, actually I used a cool tool by @ZviBaratz that found the issue and it was simple enough to fix. If you find any issues with your feature on the `gnome-50` branch, you can open another PR and point it to `gnome-50` instead of master.

PR #577, opened 2025-12-30
Guard against updates after destroy

Pappmann:
## Summary

  Prevent async callbacks and UI updates from touching destroyed widgets after the extension is disabled, which can trigger
  “object disposed” errors.

  ## Changes

  - Track a _destroyed flag on ClipboardIndicator.
  - Early‑return in _buildMenu, _refreshIndicator, and menu helper methods if destroyed.
  - Skip post‑build initialization when teardown already happened.

  ## Testing

  Manual: reloaded the extension and exercised clipboard history and menu.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -73,6 +73,7 @@ const ClipboardIndicator = GObject.registerClass({
     #refreshInProgress = false;
 
     destroy () {
+        this._destroyed = true;
         this._disconnectSettings();
         this._unbindShortcuts();
         this._disconnectSelectionListener();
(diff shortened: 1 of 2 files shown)


Tudmotu:
Thank you for this contribution @Pappmann !
It is planned to be included in the next version, but since e.g.o has a large backlog, it might take a while before I submit it.
Currently it is included in the `gnome-50` branch.

Tudmotu:
Thank you for this PR. Closing manually. Github doesn't close automatically because there were changes added after the original merge.

PR #578, opened 2026-01-07
Updated German translation and .pot file

normanhenges:
Updated the German translation.

Also updated the .pot file and re-generated all other .po files based on this new template.


Tudmotu:
@normanhenges Thank you for this contribution.

PR #582, opened 2026-01-14
Updated cs translation, updated .pot file

lev741:
Updated .pot file because outdated


Tudmotu:
@lev741 Thank you for this contribution.

PR #583, opened 2026-01-20
Notification setting on clear history

whilenot-dev:
This PR does two things:

1. It renames the setting `Show confirmation on Clear History` to `Prompt for confirmation on Clear History` and moves it to the UI-section. image
2. It introduces a new Notifications-setting `Show notification on Clear History` which has the current behavior as default (notification is shown). image


Tudmotu:
Thank you @whilenot-dev for this PR. I believe your fix also made me realize I had a bug in the settings binding, possibly related to #587.

I plan on releasing this change in the next version, but due to large backlog on e.g.o it might take a while before the next version can be approved. For now, you can use branch `gnome-50` if you want to use your feature.

Issue #502, opened 2025-01-04
Extension stops working due to invalid cache file


pedrovhb:
Hey! First of all, thanks for the extension, it's quite helpful :)

Found an issue and workaround, but thought I'd make you aware of it -

One day, the extension stopped working for me. Now after having investigated, I found that `journalctl -f` reported the following when copying something:

    jan 04 07:12:07 pc .gnome-shell-wr[28828]: JS ERROR: SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data @ file:///home/pedro/.local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com/registry.js:92:51
                                               read/</</<@file:///home/pedro/.local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com/registry.js:92:51
                                               @resource:///org/gnome/shell/ui/init.js:21:20

I figure something got overwritten with an empty file when it was meant to contain JSON, and trying to parse this breaks everything. Looking at the code I found the cache dir, and indeed doing `rm -rf ~/.cache/clipboard-indicator@tudmotu.com` fixed the issue.

Not sure if the invalid cache originated from the extension itself or if I inadvertently messed with it in some other way. We'd probably want to catch the parsing error and have the extension remove the file when this happens.

markovic-nikola:
Happened to me too, it was a very large JSON file.

Tudmotu:
Not sure what caused the JSON to break but I suspect the file size limit might be related. I'm marking as a bug, if the file size limit is indeed the culprit then the solution would probably be to remove the file size limit.

aronwolf90:
@pedrovhb thanks a lot! removing the file solved the issue on my side.

amirsobhi:
removing the cache deletes all the pinned text. Not a solution I'm looking for.
So far, I have been disabling and enabling it to get it working (using tweaks).

anirudhsanthosh:
Happened to me too. when I cleared all the history.

PR #586, opened 2026-02-13
Handle the empty registry cache file

mxchist:
I ran into the same issue as described in #502. In my case, the `JS ERROR: SyntaxError: JSON.parse: unexpected end of data at line 1 column 1` was caused by an empty registry file (registry.txt).

Unfortunately, the issue appears to have started quite some time ago (likely in the second half of 2024). I postponed  investigating why the extension stopped working until this week, and by now I can no longer determine what originally caused the registry file to be emptied.

### DIFF
--- a/registry.js
+++ b/registry.js
@@ -89,7 +89,15 @@ export class Registry {
 
                         if (success) {
                             let max_size = this.settings.get_int(PrefsFields.HISTORY_SIZE);
-                            const registry = JSON.parse(new TextDecoder().decode(contents));
+                            const cacheTextData = new TextDecoder().decode(contents);
+                            var registry;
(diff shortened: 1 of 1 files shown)


mxchist:
I attempted to make the registry file empty by:
* pressing the 🗑️ Clear history button;
* manually deleting all history entries one by one;
* enabling the "Clear history on system reboot" option.
None of these actions resulted in an empty cache file; therefore, the file never became invalid input for JSON.parse.

Tudmotu:
Thanks @mxchist. This change seems harmless to me so I plan on merging it, but it will still take a while since e.g.o reviews currently have a large backlog. If you'd like to test it out with the next version of the extension, you can use the `gnome-50` branch for now.

PR #598, opened 2026-03-31
Fix menu builder object references

JaredTweed:
Replaces stray `that` references in `_buildMenu()` with `this`. The undefined identifier caused menu construction to fail under GNOME Shell's ES module runtime, which made the panel button appear to do nothing. Tested inside GNOME shell.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -683,6 +683,11 @@ const ClipboardIndicator = GObject.registerClass({
             menuItem.actor.add_child(menuItem.tagLabel);
         }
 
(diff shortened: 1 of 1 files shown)


Tudmotu:
Ah thank you, I accidentally pushed a search&replace without properly testing it. Will take a look in a bit.

Tudmotu:
Also thank you for fixing the alignment issue, it didn't come up in my tests.

JaredTweed:
No problem! Happy to help.

PR #603, opened 2026-04-08
feat: show confirmation dialog before deleting pinned clipboard items

MAAB-FW:
This PR addresses an issue with pinned clipboard items being unintentionally deleted.

## Problem:
When deleting clipboard entries using the "Delete" button, pinned items are also removed. This makes it difficult to reliably use the pinned clipboard feature.

## Solution:
A confirmation dialog is now shown when attempting to delete a pinned item, preventing accidental removal.

### Notes:
If this approach is not preferred, an alternative could be to add a toggle option in settings to control this behavior.


Tudmotu:
Thank you for this PR @MAAB-FW . I am willing to accept this feature but it needs to be slightly reworked:
1. You need to base your changes on branch `gnome-50`
2. You should add a settings options. Put it in the "UI" section, below "prompt for confirmation on clear history". The label should be "Prompt before deleting pinned item", and the subtitle should be "Ask for confirmation before deleting a pinned item"

MAAB-FW:
Thanks for the feedback. I've updated the settings UI and rebased it on `gnome-50`. Let me know if everything looks good now.

Tudmotu:
Thank you. Can you please make these changes as well:
1. Change the default to "false" to maintain the current behavior
2. Make sure to check the value of the setting `CONFIRM_ON_PINNED_DELETE` also when pressing `Key_Delete`

MAAB-FW:
All set. Take a look.

PR #597, opened 2026-03-30
Gnome 50 release (includes the PRs below this one)

Tudmotu:
This is the next major release. It includes the previous PRs and some other features.

Main PRs merged:
- https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/534
- https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/570
- https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/577
- https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/pull/583

Other features:
- Image preview
- Item editing
- Item labelling
- Setting to open menu at cursor (experimental i.e. buggy)
- Setting to hide the item action buttons

PR #620, opened 2026-05-22
Fix memory bloat, crash on corrupt cache, and 3 bugs

### LINKED ISSUE #502: Extension stops working due to invalid cache file

### LINKED ISSUE #559: "Paste on select" breaks "Move item to top after selection"
I swear this used to work because it's my preferred behavior and it stopped somewhat recently, but I don't know when.

When both "Paste on select" and "Move item to top after selection" are enabled, the selected item is pasted but not moved to the top, which is unexpected, I want it to both paste and move the pasted/selected item to the top.

Gnome 48 on Fedora 42, extension version 69 (but it also did not work on version 68).

### LINKED ISSUE #587: Clear history notification is shown when disabled
Hi,
I am using "Clear clipboard history on interval" and since I do not want to get a notification for that I also disabled all three notification settings. But it seems I am still getting the notification that the clipboard history has been cleared on every interval.

Is this intended or a bug?

Steps to reproduce:
1. Enable "Clear clipboard history on interval"
2. Set "Clear history interval (in minutes)" to a value of your choice (1 min for impatient people)
3. Disable all notification settings
4. Wait for the set cycle time

### LINKED ISSUE #589: The extension has stopped functioning due to malformed JSON in the cache file.
What happened: The extension no longer displays copied items in the history, and its icon does not respond to mouse clicks.  
What is expected: The extension should store the copied entities correctly.
The message in the syslog:
```
gnome-shell[1915]: JS ERROR: SyntaxError: JSON.parse: unexpected non-whitespace character after JSON data at line 1 column 1681 of the >
                                         read/</</<@file://${HOME}/.local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com/registry.js:99:49
                                         @resource:///org/gnome/shell/ui/init.js:21:20
```

Current state of the JSON in the cache file: You can view the cache file here https://pastebin.com/6Tz2xhmN, password is YYE00FWxRG.

What happened before the issue appeared: I reinstalled the Linux system while retaining the existing /home partition, which includes the Clipboard Indicator extension and its data.

The timeline of events is as follows:  https://pastebin.com/HMz9ZcZ4, password: YYE00FWxRG.

The current opetating system is Nixos 25.11, Gnome v.49, Clipboard Indicator v.69.
The settings: ${HOME}/local/share/gnome-shell/extensions/clipboard-indicator@tudmotu.com/schemas/org.gnome.shell.extensions.clipboard-indicator.gschema.xml: https://pastebin.com/WQSfSved, password: YYE00FWxRG

Further context:
The entity "Церковь Святого Георгия" was previously copied with an extra whitespace at the end, and on another occasion without it. This variation likely caused a partial concatenation within the JSON array, rendering it invalid.

### LINKED ISSUE #613: "Move the item to the top after selection" does not work properly when "paste on select" is enabled.
**Bug Describtion**
After enabling the "paste on select" option, when you select any option in the clipboard, the selected option moves to the second line of the clipboard.

**To Reproduce**
Enable both "paste on select" and "Move the item to the top after selection", then select any option except the first line.

Image

**Expected behavior**
After selection, it should move to the top line.

**Device information**
OS: Fedora 43
Kernel: x86_64 Linux 6.19.14-200.fc43.x86_64

**Affected  extension versions**
Newest version in https://github.com/Tudmotu/gnome-shell-extension-clipboard-indicator/tree/master 、v71

**Screenshots & Attachments**

Image

### LINKED ISSUE #617: It freezes the system when there are too many records
After some time, when I reached about 700 items (mixed content of images and text) in cache folder (.cache/clipboard-indicator@tudmotu.com) every attempt to change settings, or remove cache from GUI ended with completely freezed system, except mouse/touchpad (hard restart was required to do anything).

I noticed that for last few days copy (CLTR+C) operation became slower and slower and I didn't know why, and it was probably because of the extension reaching it's limits.

Info:

I had a limit 1500 items, and 500MB cache.

Extension version: 71

OS: Ubuntu 26.04 resolute
 Kernel: x86_64 Linux 7.0.0-15-generic
 Uptime: 16m
 Packages: 2521
 DE: GNOME 50.0
 WM: Mutter
 Disk: 614G / 915G (71%)
 CPU: 13th Gen Intel Core i5-1335U @ 12x 4.6GHz [42.0°C]
 GPU: Mesa Intel(R) Iris(R) Xe Graphics (RPL-U)
 RAM: 5822MiB / 14994MiB

leogallego:
## Summary

This PR addresses several open issues:

- **Lazy-load image bytes at startup**: image entries now store only a file path and hash instead of reading full bytes into memory. Bytes are loaded on demand when pasting, then released. Eliminates the startup memory spike described in #617 (hundreds of MB of images loaded into gnome-shell heap).
- **Clean orphaned cache files on startup**: unreferenced image files in `~/.cache/clipboard-indicator@tudmotu.com/` are deleted when the extension loads.
- **Handle malformed JSON in cache gracefully** (#589, #502): corrupted `registry.txt` no longer kills the extension — it backs up the bad file and starts with empty history.
- **Fix move-to-top with paste-on-select** (#613, #559): the pasted item now correctly moves to position 1 instead of position 2. Root cause was `#pasteItem` restoring the previous clipboard selection after 100ms, clobbering the `_moveItemFirst` reorder.
- **Suppress auto-clear notification** (#587): automatic interval clears no longer show a notification — only manual "Clear history" does.
- **Fix unhandled error in _openSettings**: guard against `openSettings()` returning undefined in nested sessions.

Each fix is a separate commit for cherry-picking.

## Issues closed

- Closes #617
- Closes #589
- Closes #502
- Closes #613
- Closes #559
- Closes #587

## Test plan

- [x] `make` compiles schemas and locales
- [x] Extension loads in `make nested-session` with zero JS errors
- [x] Corrupted `registry.txt` — extension logs error, backs up bad file, starts with empty history (verified in nested session)
- [x] 49 orphaned cache files removed on startup (verified in nested session)
- [x] Copy/paste text and images works
- [x] Settings button works without JS errors in nested session
- [x] "Paste on select" + "Move item to top" — selected item moves to position 1
- [x] "Clear history on interval" with notifications off — no notification on auto-clear

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -1636,11 +1636,12 @@ const ClipboardIndicator = GObject.registerClass({
 
 
 
-    #pasteItem (menuItem) {
+    async #pasteItem (menuItem) {
         this.menu.close();
         const currentlySelected = this._getCurrentlySelectedItem();
+        const shouldRestore = !PASTE_ON_SELECT && currentlySelected && currentlySelected !== menuItem;
         this.preventIndicatorUpdate = true;
-        this.#updateClipboard(menuItem.entry);
+        await this.#updateClipboard(menuItem.entry);
         this._pastingKeypressTimeout = setTimeout(() => {
             if (this.keyboard.purpose === Clutter.InputContentPurpose.TERMINAL) {
                 this.keyboard.press(Clutter.KEY_Control_L);
@@ -1657,10 +1658,10 @@ const ClipboardIndicator = GObject.registerClass({
                 this.keyboard.release(Clutter.KEY_Shift_L);
             }
 
-            this._pastingResetTimeout = setTimeout(() => {
+            this._pastingResetTimeout = setTimeout(async () => {
                 this.preventIndicatorUpdate = false;
-                if (currentlySelected && currentlySelected.entry)
-                    this.#updateClipboard(currentlySelected.entry);
+                if (shouldRestore && currentlySelected.entry)
+                    await this.#updateClipboard(currentlySelected.entry);
             }, 50);
         }, 50);
     }
--- a/extension.js
+++ b/extension.js
@@ -406,6 +406,12 @@ const ClipboardIndicator = GObject.registerClass({
             this._selectMenuItem(clipItemsArr[lastIdx]);
         }
 
+        // Clean up orphaned image files in the cache directory
+        const referencedPaths = new Set(
+            clipHistory.filter(e => e.isImage() && e.getFilePath()).map(e => e.getFilePath())
+        );
+        this.registry.cleanOrphanedFiles(referencedPaths);
+
         this.#showElements();
     }
 
@@ -926,7 +932,7 @@ const ClipboardIndicator = GObject.registerClass({
         }
     }
 
-    _onMenuItemSelected (menuItem, autoSet) {
+    async _onMenuItemSelected (menuItem, autoSet) {
         for (let otherMenuItem of menuItem.radioGroup) {
             let clipContents = menuItem.clipContents;
 
@@ -935,7 +941,7 @@ const ClipboardIndicator = GObject.registerClass({
                 if (menuItem._ornamentIcon) menuItem._ornamentIcon.opacity = 255;
                 menuItem.currentlySelected = true;
                 if (autoSet !== false)
-                    this.#updateClipboard(menuItem.entry);
+                    await this.#updateClipboard(menuItem.entry);
             }
             else {
                 otherMenuItem.setOrnament(PopupMenu.Ornament.DOT);
@@ -945,12 +951,12 @@ const ClipboardIndicator = GObject.registerClass({
         }
     }
 
-    _selectMenuItem (menuItem, autoSet) {
-        this._onMenuItemSelected(menuItem, autoSet);
+    async _selectMenuItem (menuItem, autoSet) {
+        await this._onMenuItemSelected(menuItem, autoSet);
         this.#updateIndicatorContent(menuItem.entry);
     }
 
-    _onMenuItemSelectedAndMenuClose (menuItem, autoSet) {
+    async _onMenuItemSelectedAndMenuClose (menuItem, autoSet) {
         for (let otherMenuItem of menuItem.radioGroup) {
             let clipContents = menuItem.clipContents;
 
@@ -959,7 +965,7 @@ const ClipboardIndicator = GObject.registerClass({
                 if (menuItem._ornamentIcon) menuItem._ornamentIcon.opacity = 255;
                 menuItem.currentlySelected = true;
                 if (autoSet !== false)
-                    this.#updateClipboard(menuItem.entry);
+                    await this.#updateClipboard(menuItem.entry);
             }
             else {
                 otherMenuItem.setOrnament(PopupMenu.Ornament.DOT);
@@ -1234,7 +1240,12 @@ const ClipboardIndicator = GObject.registerClass({
     }
 
     _openSettings () {
-        this.extension.openSettings();
+        try {
+            const result = this.extension.openSettings();
+            if (result?.catch) result.catch(e => console.error(e));
+        } catch (e) {
+            console.error(e);
+        }
         this.menu.close();
     }
 
@@ -1555,8 +1566,8 @@ const ClipboardIndicator = GObject.registerClass({
     _selectEntryWithDelay (entry) {
         this._selectMenuItem(entry, false);
 
-        this._delayedSelectionTimeoutId = setTimeout(() => {
-            this._selectMenuItem(entry);  //select the item
+        this._delayedSelectionTimeoutId = setTimeout(async () => {
+            this._selectMenuItem(entry);
             this._delayedSelectionTimeoutId = null;
         }, DELAYED_SELECTION_TIMEOUT);
     }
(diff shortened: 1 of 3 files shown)


extension.js
```diff
@@ -871,11 +877,8 @@ const ClipboardIndicator = GObject.registerClass({
             }
         });
 
-        if (NOTIFY_ON_CLEAR) {
-            const message = invokedAutomatically
-                ? _("Clipboard history cleared automatically")
-                : _("Clipboard history cleared");
-            this._showNotification(message);
+        if (NOTIFY_ON_CLEAR && !invokedAutomatically) {
+            this._showNotification(_("Clipboard history cleared"));
```
whilenot-dev, line 881: This change doesn't seem right to me. Users might still want to get a notification, even if the history is cleared on an interval (automatically).

This setting here

Image

toggles the variable `NOTIFY_ON_CLEAR`. It's set to `true` by default.

I think it's related to #587, but that issue should have been fixed already with the GNOME 50 release #597.

extension.js
```diff
@@ -1555,8 +1563,8 @@ const ClipboardIndicator = GObject.registerClass({
     _selectEntryWithDelay (entry) {
         this._selectMenuItem(entry, false);
 
-        this._delayedSelectionTimeoutId = setTimeout(() => {
-            this._selectMenuItem(entry);  //select the item
+        this._delayedSelectionTimeoutId = setTimeout(async () => {
+            this._selectMenuItem(entry);
             this._delayedSelectionTimeoutId = null;
         }, DELAYED_SELECTION_TIMEOUT);
```
whilenot-dev, line 1569: This needs some further work I'm afraid, as timeouts don't play well with promises (`async` functions are [colored functions](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/)).

First off, we shouldn't loose references to dangling promises, so we don't accidentally crash the runtime. You got (at least) two dangling promises here in your changes:

1. `this._selectMenuItem(entry, false);` on line 1564
1. `this._selectMenuItem(entry);` on line 1567

A thrown error in either one will bring down the whole runtime and if we don't `await` these promises, we can't even make use of a `try`-scope to catch these. 

Further, and that might be the strongest point to prefer timeouts over promises, we can cancel timeouts but we can't cancel promises in JavaScript (\*we can let an unwanted promise do nothing, but more on that further down).

This whole function should probably be rewritten to something analog to this, which would make `_selectEntryWithDelay` `async` as well (and add color to it):

```js
async _selectEntryWithDelay (entry) {
  await this._selectMenuItem(entry, false);
  await delay(DELAYED_SELECTION_TIMEOUT);
  await this._selectMenuItem(entry);
}

// util
function delay(ms) {
  return new Promise((resolve) => {
    setTimeout(resolve, ms);
  });
}
```

That brings us to another problem though... We currently use timeout IDs as synchronization primitives and we use these IDs to cancel asynchronous procedures. As I wrote earlier, "we can let an unwanted promise do nothing", but that inverts the responsibility of cancellation.

What was with timeouts:

```js
// foo.js
let state = false;

// cancel handling
let timeoutId = null;
setTimeout(() => {
  if (timeoutId !== null) {
    clearInterval(timeoutId);
  }
}, 500);

// state update
timeoutId = setTimeout(() => {
  state = true;
  timeoutId = null;
}, 1000);
```

now becomes with promises:

```js
// bar.js
let state = false;

// cancel handling
let isCancelled = false;
const cancel = (async () => {
  await delay(500);
  isCancelled = true;
})();
 
// state update
const update = (async () => {
  await delay(1000);

  // cancellation handling (inside async callback)
  if (isCancelled) {
    return;
  }

  state = true;
})();

await Promise.all([
  cancel,
  update,
]);
```

I hope I made myself clear: The (simulated) cancellation of promises requires additional state handling inside the async callback, meaning the async callback implementation needs to be aware of the used synchronization primitives (the `isCancelled` flag in this case) and branch accordingly after each `await`ed promise. **<- this is the further work required**

If you run both examples in your JS-runtime of choice, you'll notice that `foo.js` exits around 500ms earlier than `bar.js`.

